PR #17, opened 2023-04-28
add ADR about copying other CAS projects

Harriethw:
We've made a lot of decisions about what we're picking and choosing so far from the other CAS projects, this ADR is a way to keep track of those decisions.


Floppy (approved):

Floppy:
All good, just needs the title change I've suggested above.

Floppy (approved):

PR #19, opened 2023-05-16
Get risk data

Harriethw:


server/controllers/peopleController.ts
```diff
@@ -0,0 +1,17 @@
+import type { Request, RequestHandler, Response } from 'express'
+import paths from '../paths/apply'
+
+export default class PeopleController {
+  find(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { crn } = req.body
+
+      if (crn) {
+        // TODO: error handling if invalid/not found CRN
```
pezholio, line 10: One thing worth pointing out (and something we need to handle in some cases) is if the user can't access the CRN (it's not it their caseload, or it's a restricted access offender (usually someone famous etc), the 403 response will bubble up from the API and log the user our.

server/data/personClient.test.ts
```diff
@@ -0,0 +1,71 @@
+import PersonClient from './personClient'
+import { oasysSectionsFactory } from '../testutils/factories'
+import paths from '../paths/api'
+
+import describeClient from '../testutils/describeClient'
+
+describeClient('PersonClient', provider => {
+  let personClient: PersonClient
+
+  const token = 'token-1'
```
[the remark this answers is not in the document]
pezholio: We got these alerts from CodeQL too when we set it up, you can safely ignore them

server/utils/formUtils.ts
```diff
@@ -0,0 +1,7 @@
+import * as nunjucks from 'nunjucks'
+
+// eslint-disable-next-line import/prefer-default-export
```
pezholio, line 3: I think we disabled this globally, it's an annoyance, especially when you know you're going to add stuff later

server/testutils/factories/oasysSections.ts
```diff
@@ -20,10 +20,10 @@ export default Factory.define<OASysSections>(() => ({
...
   questionNumber: options.sequence.toString(),
   label: faker.helpers.arrayElement([
-    'Who is at risk?',
-    'What is the nature of the risk?',
-    'When is the risk likely to be greatest?',
-    'What circumstances are likely to increase risk?',
+    'Who is at risk',
+    'What is the nature of the risk',
+    'When is the risk likely to be greatest',
+    'What circumstances are likely to increase risk',
```
pezholio: Can these changes be rolled into the commit when you created the factories? I don't think they belong here

pezholio (dismissed):
LGTM - I'm a big fan of pulling stuff in as and when you need it, rather than copying it all over wholesale. I'd really like to break a lot of the stuff we share (particularly the utils) into a shared NPM package or something, but time!

PR #26, opened 2023-05-23
Feature/error handling

patrickjfl:
## Summary of changes

This PR adds error handling for the CRN search feature, so that users will be notified if they do not enter a CRN, or enter a CRN that does not exist, or that they are not authorised to access. Any other error will be thrown and handled generically.

A lot of the code outside the peopleController has been copied over from CAS-1 with some minor modifications (e.g. removing unnecessary function parameters).

## Screenshots

Screenshot 2023-05-23 at 10 40 59

Screenshot 2023-05-23 at 11 20 12

### DIFF
--- a/server/controllers/peopleController.test.ts
+++ b/server/controllers/peopleController.test.ts
@@ -3,8 +3,12 @@ import { DeepMocked, createMock } from '@golevelup/ts-jest'
 
 import paths from '../paths/apply'
 import PeopleController from './peopleController'
+import { errorMessage, errorSummary } from '../utils/validation'
+import PersonService from '../services/personService'
+import personFactory from '../testutils/factories/person'
 
-describe('applicationsController', () => {
+describe('peopleController', () => {
+  const flashSpy = jest.fn()
   const token = 'SOME_TOKEN'
   const crn = '1234'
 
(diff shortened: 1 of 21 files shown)


server/controllers/peopleController.ts
```diff
@@ -1,17 +1,43 @@
...
+        } catch (err) {
+          if (err.status === 404) {
+            this.addErrorMessagesToFlash(req, `No person with a CRN of '${crn}' was found`)
+          } else if (err.status === 403) {
+            this.addErrorMessagesToFlash(req, 'You do not have permission to access this CRN')
+          } else {
+            throw err
+          }
+
+          res.redirect(req.headers.referer || '')
```
Harriethw, line 27: this is my bad, I left in the empty string earlier because typescript was complaining, we can remove it now (and the test that checks for it in `peopleController.test.ts`
```suggestion
          res.redirect(req.headers.referer)
```

server/controllers/peopleController.test.ts
```diff
@@ -3,39 +3,127 @@ import { DeepMocked, createMock } from '@golevelup/ts-jest'
...
+
+        await requestHandler(request, response, next)
+
+        expect(response.redirect).toHaveBeenCalledWith('some-referrer/')
+
+        expect(flashSpy).toHaveBeenCalledWith('errors', { crn: errorMessage('crn', 'You must enter a CRN') })
+        expect(flashSpy).toHaveBeenCalledWith('errorSummary', [errorSummary('crn', 'You must enter a CRN')])
       })
 
       it('does not redirect if there is not a referrer url', () => {
```
Harriethw, line 129: as above comment, this test can be deleted

Harriethw (approved):
Nice one! Just those changes with the `referer` and then it's good to go 🚀

server/controllers/peopleController.ts
patrickjfl: ah nice one, thanks

PR #36, opened 2023-06-01
rename `supported-accommodation` to `community-accommodation-tier-2`

Harriethw:
We have decided to use CAS-2 (Community Accommodation Service Tier 2) as the main user-facing name for the service, so we are renaming the repo. 

It doesn't seem that historically we put `service` in a repo name, so I've left that out, and used the full words rather than acronym to avoid another acronym on top of `hmpps`.


patrickjfl (approved):
LGTM 👍

PR #43, opened 2023-06-06
214 save an application

patrickjfl:
This PR allows users to save an application, as well as rendering a list of existing applications to the user on the 'New applications' screen.

Screenshot 2023-06-06 at 16 28 07

### DIFF
--- a/integration_tests/tests/apply/new.cy.ts
+++ b/integration_tests/tests/apply/new.cy.ts
@@ -1,8 +1,10 @@
+import { Cas2Application as Application } from '@approved-premises/api'
 import IndexPage from '../../pages'
-import { personFactory } from '../../../server/testutils/factories/index'
+import { personFactory, applicationFactory } from '../../../server/testutils/factories/index'
 
-context('Risks', () => {
+context('New', () => {
   const person = personFactory.build({})
+  const applications = applicationFactory.buildList(3) as Array<Application>
 
   beforeEach(() => {
     cy.task('reset')
@@ -20,15 +22,22 @@ context('Risks', () => {
         oasysSections: riskData,
       })
     })
+    // and existing applications
+    cy.task('stubApplications', applications)
   })
 
-  it('allows user to enter a CRN and see the first risk question data', () => {
+  it('shows existing applications, allows user to enter a CRN and see the first risk question data', () => {
     // Given I visit the start page
     IndexPage.visit()
     // And I click on the start button
     cy.get('[role="button"]').click()
 
-    // Then I see the CRN form
+    // Then I see all existing applications
+    applications.forEach(app => {
+      cy.get('td').contains(app.person.crn)
+    })
+
+    // And I see the CRN form
     // And enter a valid CRN
     cy.task('stubFindPerson', {
       person,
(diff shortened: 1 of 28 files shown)


integration_tests/tests/apply/new.cy.ts
```diff
@@ -61,7 +70,12 @@ context('Risks', () => {
     // And I click on the start button
     cy.get('[role="button"]').click()
 
-    // Then I see the CRN form
+    // Then I see all existing applications
```
Harriethw, line 73: minor, but I don't think we need to check this again in this test, as it's not the functionality under test and the circumstances haven't changed from the first one.

integration_tests/tests/apply/risk.cy.ts
```diff
@@ -0,0 +1,39 @@
...
+      const application = applicationFactory.build({})
+      cy.task('stubCreateApplication', { application })
+
+      // given I am on the risks page
+      cy.visit(paths.applications.show({ crn: person.crn }))
+
+      // when I click the save and continue button
+      cy.get('button').click()
+
+      // then I should be redirected back to the new page
```
Harriethw, line 36: are we missing this, or are we able to check that it's gone to the right page?

server/views/applications/new.njk
```diff
@@ -13,8 +14,30 @@
...
+                  firstCellIsHeader: true,
+                  head: [
+                    {
+                      text: "Person"
+                    },
+                    {
+                      text: "CRN"
+                    }
+                  ],
+                  rows: dashboardTableRows(applications.applications)
```
Harriethw, line 33: should we just pass in `applications` instead of `applications.applications`? I'm getting an error atm...

integration_tests/tests/apply/new.cy.ts
```diff
@@ -20,15 +22,17 @@ context('Risks', () => {
         oasysSections: riskData,
       })
     })
+    // and existing applications
+    cy.task('stubApplications', { applications })
```
Harriethw, line 26:
```suggestion
    cy.task('stubApplications', applications)
```

think we just pass in the list here rather than an object

integration_tests/tests/apply/risk.cy.ts
patrickjfl: good spot - I'm just adding this

server/data/applicationClient.ts
```diff
@@ -0,0 +1,23 @@
...
+  restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('applicationClient', config.apis.approvedPremises as ApiConfig, token)
+  }
+
+  async create(crn: string): Promise<Application> {
+    return (await this.restClient.post({
+      path: paths.applications.new.pattern,
+      data: { crn: crn.trim() },
```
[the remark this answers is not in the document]
patrickjfl: I think I actually copied this from CAS3 - it was introduced there as a temporary measure until the Kotlin devs had time to look at it.

integration_tests/tests/apply/new.cy.ts
patrickjfl: ah yes, and I think that fixes the issue you raised above. Thanks!

Harriethw (approved):

PR #70, opened 2023-07-03
Sentry integration

Harriethw:
We will be using Sentry to track errors in the app. We're following the
[same implementation as CAS1 + 2](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/commit/f4fd6ce072799d559d13c307a05229f35ac19ed7), with slight difference that we do not
[need to use the `@sentry/tracing` package because it is being
deprecated](https://docs.sentry.io/product/sentry-basics/tracing/).

I've tested this out locally by creating an error [and it has been logged here](https://ministryofjustice.sentry.io/issues/4293405489/?project=4505442489729024&referrer=project-issue-stream)

### DIFF
--- a/server/middleware/setUpSentry.ts
+++ b/server/middleware/setUpSentry.ts
@@ -1,0 +1,63 @@
+import * as Sentry from '@sentry/node'
+import express from 'express'
+import applicationVersion from '../applicationVersion'
+import config from '../config'
+
+export function setUpSentryRequestHandler(app: express.Express): void {
+  if (config.sentry.dsn) {
+    // Prevent usernames which are PII from being sent to Sentry
+    // https://docs.sentry.io/platforms/python/guides/logging/data-management/sensitive-data/#examples
+    const anonymousId = Math.random().toString()
+    Sentry.setUser({ id: anonymousId })
+
+    Sentry.init({
+      dsn: config.sentry.dsn,
+      environment: config.environment,
+      release: applicationVersion.gitRef,
+      integrations: [
+        // enable HTTP calls tracing
+        new Sentry.Integrations.Http({ tracing: true }),
+        // enable Express.js middleware tracing
+        new Sentry.Integrations.Express({ app }),
+      ],
+      tracesSampler: samplingContext => {
+        const transactionName = samplingContext?.transactionContext?.name
+        if (
+          transactionName?.includes('ping') ||
+          transactionName?.includes('health') ||
+          transactionName?.includes('assets')
+        ) {
+          return 0
+        }
+
+        if (config.environment === 'prod') {
+          return 1
+        }
+
+        // Default sample rate
+        return 0.05
+      },
+      beforeSend(event) {
+        if (event.user) {
+          // Don't send username
+          // eslint-disable-next-line no-param-reassign
+          delete event.user.username
+        }
+        return event
+      },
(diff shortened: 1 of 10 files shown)


server/middleware/setUpSentry.ts
```diff
@@ -0,0 +1,56 @@
...
+import express from 'express'
+import applicationVersion from '../applicationVersion'
+import config from '../config'
+
+export function setUpSentryRequestHandler(app: express.Express): void {
+  if (config.sentry.dsn) {
+    // Prevent usernames which are PII from being sent to Sentry
+    // https://docs.sentry.io/platforms/python/guides/logging/data-management/sensitive-data/#examples
+    const anonymousId = Math.random().toString()
+    Sentry.setUser({ id: anonymousId, username: anonymousId })
```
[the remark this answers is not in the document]
Harriethw: we are using this to hide usernames rather than security, so this should be fine

integration_tests/e2e/health.cy.ts
```diff
@@ -27,4 +27,13 @@ context('Healthcheck', () => {
       })
     })
   })
+
+  context('Sentry check', () => {
+    it('Throws a test error', () => {
+      cy.request({ url: '/debug-sentry', method: 'GET', failOnStatusCode: false }).then(response => {
+        expect(response.status).to.equal(500)
+        expect(response.body).to.contain('Test Sentry error thrown.')
```
edavey, line 35: Nice!

Would it be worth including a test here to demonstrate that the PII are being anonymised as expected?

As far as I can tell we aren't actually able to run this test at present? I've seen [this PR](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/compare/220-e2e-tests) about using Playwright but I think that is a work in progress?

server/middleware/setUpSentry.ts
```diff
@@ -0,0 +1,55 @@
...
```
Gweaton, line 11: Unfortunately this didn't prevent the username being sent as a "tag" for us - I've merged another PR that appears to have fixed it [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/99) so on the off chance this doesn't work for you either, you may want to follow our example

Gweaton (approved):

server/middleware/setUpSentry.ts
Harriethw: argh, good catch, it is sharing the local username 😱

integration_tests/e2e/health.cy.ts
Harriethw: > As far as I can tell we aren't actually able to run this test at present?

ah yeah that's a good point 🤦 I've realised this is in the `e2e` folder, but I think it can be run as a regular integration test...

> Would it be worth including a test here to demonstrate that the PII are being anonymised as expected?

That's a really good idea!

Harriethw: ah just actually looking at the context of this test, because I realised it wasn't really testing anything... it looks like it was put in to force a production error to be thrown in order for the environment to be seen https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/commit/2acadacbbab80a498c6675b3de905011a4b7d6d1

PR #97, opened 2023-07-21
Opening in progress applications 

Harriethw:
We would like users to be able to open the applications they have started.
This PR:
* renames the `new` applications page to be the `index` (there will be a new `new` page created, where the CRN form will move to)
* adds a link to the application on the applicant's name.


integration_tests/pages/apply/list.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  private shouldShowApplications(applications: Array<Cas2ApplicationSummary>): void {
+    applications.forEach(application => {
+      cy.contains(application.person.name)
+        .should('have.attr', 'href', paths.applications.show({ id: application.id }))
+        .parent()
+        .parent()
+        .within(() => {
+          cy.get('th').eq(0).contains(application.person.name)
```
patrickjfl, line 31: Do we gain anything from this check? Seems to me like we've already checked this on line 26...

server/utils/taskListUtils.test.ts
```diff
@@ -67,5 +67,13 @@ describe('taskListUtils', () => {
         '<strong class="govuk-tag govuk-tag--grey app-task-list__tag" id="second-task-status">Cannot start yet</strong>',
       )
     })
+
+    it('returns a completed tag when the task cannot be started', () => {
```
patrickjfl, line 71: I think the test name is wrong here (and also on line 47, by the looks)

patrickjfl: or in fact, it looks like the test on line 47 should be deleted

patrickjfl (approved):
Looks good to me! Just a couple of minor comments 👍

integration_tests/pages/apply/list.ts
Harriethw: ah really good spot. I've now copied CAS3's way of doing this which I think is a bit better and more precise:
``` 
cy.get(`a[href*="${paths.applications.show({ id: application.id })}"]`)
        .parent()
        .parent()
        .within(() => {
          cy.get('th').eq(0).contains(application.person.name)
        })
 ```

patrickjfl: looks good!

PR #100, opened 2023-07-24
Add `test` environment configuration

Harriethw:
We would like to create a test environment for CAS2 so that we can easily demo features.
The [bootstrap](https://github.com/ministryofjustice/dps-project-bootstrap/pull/468) and [cloud PRs](https://github.com/ministryofjustice/cloud-platform-environments/pull/14813) have been merged.
Like CAS1 and 3, deployment to test will be manual, so that we have control over what features are released there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,26 @@ workflows:
                 - main
           requires:
             - deploy_dev
+      - request-test-approval:
+          type: approval
+          requires:
+            - deploy_dev
+            - e2e_environment_test
+      - hmpps/deploy_env:
+          name: deploy_test
+          env: 'test'
+          jira_update: true
+          context:
+            - hmpps-common-vars
+            - hmpps-community-accommodation-tier-2-ui-stage
+          filters:
+            branches:
+              only:
+                - main
+          requires:
+            - helm_lint
+            - build_docker
+            - request-test-approval
   #      - request-preprod-approval:
   #          type: approval
   #          requires:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -121,6 +121,26 @@ workflows:
...
+          requires:
+            - deploy_dev
+            - e2e_environment_test
+      - hmpps/deploy_env:
+          name: deploy_test
+          env: 'test'
+          jira_update: true
+          context:
+            - hmpps-common-vars
+            - hmpps-community-accommodation-tier-2-ui-stage
```
edavey, line 135: I'm wondering why this is 'context' is known as 'stage'? Wouldn't 'test' be more consistent and clear?

edavey (approved):
One little query over the term 'stage' which seems at odds with the concept of 'test'? But probably I'm misunderstanding something...

We take on significant maintenance overhead with this additional environment which no doubt will have different data, deployment and upstream connection requirements. 

I guess that if we don't see the benefit we can always drop this environment!

.circleci/config.yml
Harriethw: yep it would, that's a good point. and this reminds me that I need to ask how this context was setup in CircleCI, as afaik it doesn't exist for us

Harriethw: oh just checked CircleCi and this context does now exist for our project, I think that it's generated by moj team as all the other projects have `-stage` on them, so won't rock the boat at this point 😬

PR #101, opened 2023-07-24
"About the person" section > Equality and diversity task > 1st page (Will answer questions?)

edavey:
In this PR we start implementing the first task in the "About the person" section. This task is an "Equality and diversity" questionnaire, and the first step in that task is to determine whether the person being released from custody wishes to answer the questions. The section/task/page hierarchy is:

```
Section: About the person
  Task: Equality and diversity
    Page 1: Will answer equality questions?
```

**Note on personalisation of the section name**

The prototype design uses a personalised section name: "About Fred Smith" rather than "About the person", which is the form implemented here. This deviation is because at  present it's not obvious how to extend the "CAS task list framework" to do something like:

```ts
@Section({  
  title: `About ${application.person.name}`,  
  tasks: [AboutThePerson],  
})
```

The pragmatic approach may well to be to avoid this design until we have time to invest in supporting that capability.

## Screenshot of task list
tasklist_about_the_person

## Screenshot of the 'Will answer equality questions?' page
will_answer_equality_questions

## Structure of integration tests

In this PR we've made a deliberate attempt to separate the description of the expected behaviour of the service from the mechanics of their implementation. What this means is:

### 1. The acceptance test 
The Cypress acceptance test `complete_area_and_funding_section.cy.ts` uses the language of BDD within a number of discrete scenarios:

- **Given**: the initial set up state  
- **When**: the action whose results are under consideration  
- **Then/And**: the results which we observe

The scenarios focus on the behaviour and intention of an individual feature, which can equally be thought of as a user-story or a benefit, e.g.

```
Scenario: follow link to first task page
  - **When** I follow the link within the section
  - **Then** I'm on the expected task page
  - **And** the task list page has the expected question and answers
```

We avoid describing the detailed mechanics of the implementation (button clicks, copy on the page) to allow us to focus on the benefit being delivered, independently of the precise implementation

### 2. The test helper
That precise implementation is captured in the page's test helper `WillAnswerEqualityQuestionsPage`. Over time and through testing the implementation details are bound to change, whilst the scenario (verifying that the expected options are present) is likely to remain consistent. By separating the implementation from the description of the benefit being delivered, the maintenance of our integration tests should be clear. When the options change, we can update the relevant assertion in our helper:

```ts
hasQuestionsAndAnswers = (): void => {
  cy.get('label').contains('Yes, answer the equality questions (takes 2 minutes)')
  cy.get('label').contains('No, skip the equality questions')
}
```


edavey:
[Equality & Diversity - Guidance & Opt out (Question only) [3]](https://trello.com/c/wKp3wNYq/428-equality-diversity-guidance-opt-out-question-only)

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
```diff
@@ -0,0 +1,192 @@
...
+//    When I select an option and choose to save and continue
+//    Then I'm returned to the task list
+//    And I see that the task is now complete
+
+import Page from '../../pages/page'
+import TaskListPage from '../../pages/apply/taskListPage'
+import WillAnswerEqualityQuestionsPage from '../../pages/apply/willAnswerEqualityQuestionsPage'
+import { personFactory, applicationFactory } from '../../../server/testutils/factories/index'
+
+context('Visit area and funding section', () => {
```
Harriethw, line 42:
```suggestion
context('Visit about the person section', () => {
```
think the contexts just need updating?

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    // Then I see that an answer is required
+    const page = new WillAnswerEqualityQuestionsPage(this.application)
+    page.shouldShowErrorMessagesForFields(['willAnswer'])
+  })
+
+  // Scenario: return to task list using the back button
+  //----------------------------------------------------
+  it('takes me back to the task list page', function test() {
+    // Given I'm on the Funding information task page
```
Harriethw, line 147:
```suggestion
    // Given I'm on the Complete equality and diversity monitoring  task page
```

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
```diff
@@ -0,0 +1,192 @@
...
+    page.clickBack()
+
+    // Then I'm on the task list page
+    Page.verifyOnPage(TaskListPage)
+  })
+
+  // Scenario: task completed successfully
+  // -------------------------------------
+  it('submits the valid form', function test() {
+    // Given I'm on the Funding information task page
```
Harriethw, line 161:
```suggestion
    // Given I'm on the Complete equality and diversity monitoring task page
```

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
```diff
@@ -0,0 +1,192 @@
...
+  beforeEach(function test() {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubAuthUser')
+
+    cy.fixture('applicationData.json').then(applicationData => {
+      const application = applicationFactory.build({
+        id: 'abc123',
+        data: {
+          'area-and-funding': {
```
Harriethw, line 54: minor question: do we need to build with the `area-and-funding` data in the application, or can that be deleted?

server/form-pages/apply/about-the-person/equalityAndDiversity.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+    return errors
+  }
+
+  response() {
+    const response = {
+      [this.questions.willAnswer]: options[this.body.willAnswer],
+    }
+
+    Object.keys(response).forEach(key => {
```
Harriethw, line 57: I noticed we iterate over the keys in `fundingInformation` too, is that needed? Won't `response` here always have one key, and as it's a radio button `response[key]` will never evaluate as false at this point? 🤔

Harriethw (approved):
Thanks for the refactoring Ed, really great explanation and I think this will put us on a really solid ground going forward. Just had a couple of minor points.

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
edavey: yes, we need to put the area-and-funding task into a complete state so that the equality and diversity task in the "Not yet started" state. If not our new task will be in `Cannot yet be started`

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
edavey: thks for catching these!

server/form-pages/apply/about-the-person/equalityAndDiversity.ts
edavey: good point, well spotted, this was more blind copy-paste!

integration_tests/tests/apply/complete_about_the_person_section.cy.ts
Harriethw: ah that's interesting, I hadn't looked at the logic in the `taskListService` that makes a section `cannot_start` - we were hoping to allow the user to start any task in any order (after the first section has been completed) so sounds like we need a user story to allow that to work.

PR #107, opened 2023-07-26
Remove pre-commit hook

patrickjfl:
This removes the pre-commit hook previously added as it was causing issues integrating Husky with Tower


edavey (approved):
👍

PR #114, opened 2023-07-28
Fix section/task hierarchy

patrickjfl:
The system of defining sections alongside tasks was causing a number of issues, including tasks on the task list page being rendered with section IDs rather than task IDs, and sections being given slugs and keys within the Application data structure. Sections should not be used for URL navigation and so do not need slugs, nor are they used in the Application data structure. Instead, tasks are used in both cases.

This PR addresses that, defining sections in their own index file within a section directory. This file generates a title and a list of tasks within that section. The tasks are located within their own subdirectory, the index file of which defines a name and a slug for the task, as well as a list of pages that are contained within the task. These pages are contained within the task subdirectory.

This PR also reorganises our integration tests for Apply to make them more modular, following a section -> task -> page directory structure so that we can test each form page within its own test file.


server/form-pages/apply/area-and-funding/funding-information/index.ts
```diff
@@ -0,0 +1,11 @@
+/* istanbul ignore file */
+
+import { Task } from '../../../utils/decorators'
+import FundingInformationPage from './fundingInformation'
+
+@Task({
+  name: 'Add funding information',
+  slug: 'funding-information',
```
Harriethw, line 8: question: do you think there is another `slug` value we could give the page (`fundingInformation.ts`), so it's different from the task's slug here? `funding-source` maybe? or `how-will-they-pay`?

Harriethw (approved):
thanks Pat, this way of structuring seems clearer to me! 🙌  If it's good with @edavey I'm happy.
And if you wouldn't [mind updating our docs](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/322-applications-in-progress/doc/task_list_system.md#one-task-in-section) when you get a chance 💯

integration_tests/pages/apply/taskListPage.ts
```diff
@@ -1,10 +1,24 @@
...
+      section.tasks.forEach((task, taskIndex) => {
+        if (sectionIndex === 0 && taskIndex === 0) {
+          cy.get('h2').contains(section.title).siblings('ul').find('.app-task-list__task-name').contains(task.title)
+        } else {
+          cy.get('h2').contains(section.title).siblings('ul').find('.app-task-list__task-name').contains(task.title)
+        }
+      })
+    })
+  }
+
```
edavey, line 21: This test helper could perhaps be clearer? Two suggestions:

### 1. Naming of helper
A more descriptive name to make it clear that it's asserting that each section contains the expected tasks. Something like `shouldShowTasksWithinTheirSections()` maybe?

### 2. More readable assertion code
Probably we can make the code in the assertion more readable and maintainable? It was the fiddly nature of making this assertion about the grouping of tasks into sections which prompted me to add the `data-section_name` attribute to each section `<li>` (see https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/commit/0c450e2cbfd464fa29f88d1fc6fbb25fc7ca485d). Perhaps we could use that simpler style of code (leveraging the `within()` function to describe grouping) here in your general purpose assertion, something like:

```
shouldShowTasksWithinTheirSections = (): void => {
    Apply.sections.forEach(section => {
      cy.get(`[data-section_name="${section.title}"]`).within(() => {
        // And I see the expected SECTION
        cy.get('.app-task-list__section').contains(section.title)

        // And I see each expected TASK
        section.tasks.forEach(task => {
          cy.get('.app-task-list__task-name').contains(task.title)
        })
      })
    })
  }
```

TIDY UP
We should probably remove the now unused helpers:

- `shouldShowAreaAndFundingSection()` and
- `shouldShowAboutPersonSection()`

edavey (approved):
👍 great work, thanks for bringing more clarity here, I struggled to see a incremental improvement like this 🙇 

I've made a suggestion about the assertion which describes how tasks are grouped into their sections on the task list.

server/form-pages/apply/area-and-funding/funding-information/index.ts
patrickjfl: yes, good spot! I'm happy with `funding-source`

integration_tests/pages/apply/taskListPage.ts
patrickjfl: yes, agree that is a significant improvement, have implemented

PR #115, opened 2023-07-31
Add HMPPS Audit integration for dev environment

Harriethw:
We would like to audit our user's journey through the app, and integrate with the HMPPS Audit service.

This PR copies over CAS3's audit service and middleware. 

* [The `auditMiddleware` takes a request handler,](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/101) and returns an audited
version of that request handler. When the audited request handler is
invoked, it will invoke the original request handler, and additionally
send an audit message using our AuditService

* The `auditService` sends the audit message via the AWS SQS queue:
Modelled after the AuditService in the `make-recall-decision-ui`
project, we add an AuditService for adding messages to a configured AWS
SQS queue. We are able to pass a username through to this service [after
discussions with the Audit team](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/108/commits/e82a27de95d86f8ab57421d3b45efba84145c047) 

* The audit middleware is then passed in when we define the routes (`server/routes/utils.ts`), and an `auditEvent` can be passed through. 

* We're taking on the CAS3 naming conventions for these audit events, to follow a pattern like `{ACTION}_{AREA}_{PAGE}` e.g. `VIEW_APPLICATION_NEW` or `UPDATE_APPLICATION_PAGE`, with optional `SUCCESS/FAILURE` to indicate whether action succeeded or failed. 

You can read more about HMPPS Audit capturing here https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/4061822977/ADR+-+Audit+event+capture


server/routes/index.ts
```diff
@@ -1,32 +1,30 @@
...
+  const { peopleController } = controllers
 
-  get(paths.applications.new.pattern, applicationsController.new())
-
-  get(paths.applications.index.pattern, applicationsController.index())
-
-  get(paths.applications.show.pattern, applicationsController.show())
-
-  post(paths.applications.create.pattern, applicationsController.create())
   post(paths.applications.people.find.pattern, peopleController.find())
```
edavey, line 25: do we need to do something here to explicitly audit "find by CRN" actions or is that handled magically? e.g.

```
post(paths.applications.people.find.pattern, peopleController.find(), { auditEvent: 'FIND_PERSON_BY_CRN' })
```

edavey (approved):
👍 nice work! Just a small query about the auditing on the route for "find by CRN"...

server/routes/index.ts
Harriethw: oh good spot, we do need to manually add it I believe

PR #119, opened 2023-08-02
[315] Confirm eligibility before you start

edavey:
## Insert "Confirm eligibility" task in between "Find by CRN" and task list

In this PR we insert a task into the the applicant journey, in between "Find by CRN" and seeing the task list for the first time:

```
section: Before you start
 task: Check eligibility
   page: Check eligibility
```

### After "Find by CRN", before first sight of task list
confirm_eligibility

### Redirected to task list after confirming eligibility
before_you_start_on_tasklist

### DIFF
--- a/integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_disability_page.cy.ts
+++ b/integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_disability_page.cy.ts
@@ -23,36 +23,21 @@ context('Visit "About the person" section', () => {
     cy.task('stubAuthUser')
 
     cy.fixture('applicationData.json').then(applicationData => {
+      applicationData['equality-and-diversity-monitoring'] = { 'will-answer-equality-questions': {}, disability: {} }
       const application = applicationFactory.build({
+        id: 'abc123',
         person,
+        data: applicationData,
       })
-      application.data = applicationData
       cy.wrap(application).as('application')
-      cy.wrap(application.data).as('applicationData')
     })
   })
 
   beforeEach(function test() {
     // And an application exists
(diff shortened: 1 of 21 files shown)


integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_disability_page.cy.ts
```diff
@@ -23,36 +23,21 @@ context('Visit "About the person" section', () => {
...
-          'will-answer-equality-questions': {},
-          disability: {},
-        },
-      },
-      person,
-    })
-    cy.task('stubApplicationGet', { application })
-    cy.task('stubApplicationUpdate', { application })
+    cy.task('stubApplicationGet', { application: this.application })
+    cy.task('stubApplicationUpdate', { application: this.application })
```
edavey: thanks, this is so much clearer now!

Harriethw (approved):
thanks Ed! this is great.

>// how could I use the YesNo type to perform this test?

I'm not sure quit what you mean by this, do you mean try and check against a type at run time? it seems v readable as is

edavey:
> thanks Ed! this is great.
> 
> > // how could I use the YesNo type to perform this test?
> 
> I'm not sure quit what you mean by this, do you mean try and check against a type at run time? it seems v readable as is

yes I think that's what I had in mind but thinking about it this is probably a case of "a question which doesn't even make sense"!

edavey:
[Eligibility - Eligibility guidance (from Tasklist) [3]](https://trello.com/c/7p4Ga5aX/315-eligibility-eligibility-guidance-from-tasklist-3)

PR #125, opened 2023-08-03
Add Sexual orientation page to Equality and Diversity task

Harriethw:
This adds the sexual orientation page to the E&D task, following disability.
I've also added some bare bones instructions to follow when adding a page - I think we can update them as we go, but I found them helpful to follow so adding for future use!

Screenshot 2023-08-03 at 15 21 47

Screenshot 2023-08-03 at 15 21 13

### DIFF
--- a/integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_sexual_orientation_page.cy.ts
+++ b/integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_sexual_orientation_page.cy.ts
@@ -1,16 +1,17 @@
-//  Feature: Referrer completes 'Sex and Gender identity' question page
+//  Feature: Referrer completes 'Sexual orientation' question page
 //    So that I can complete the 'Equality questions' task
 //    As a referrer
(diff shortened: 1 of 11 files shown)


integration_tests/tests/apply/about_the_person/equality_and_diversity/complete_sexual_orientation_page.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+//    So that I can complete the 'Equality questions' task
+//    As a referrer
+//    I want to answer questions on the sex and gender page
+//
+//  Scenario: submit sex and gender itentity answers
+//    Given I'm on the 'Sex and gender identity' task page
+//    When I give valid answers to the 'Sex and gender identity' questions
+//    Then I return to the task list page
+//    And I see that the task has been completed
+
```
edavey, line 11: I think that references to "sex and gender" should be "sexual orientation"?

edavey (approved):
on 🔥

PR #130, opened 2023-08-07
Add 'Health needs' task within new 'Risks and needs' section

edavey:
## Read 'Health needs' guidance

See [Trello 328](https://trello.com/c/rmEOYpyy/328-health-needs-guidance-page)

The first task to be implemented in the "Risks and needs" section is 'Health needs'. The first page of this task is a guidance page which doesn't ask any questions but will (in a later PR) offer some copy-able "Get in touch" content for use in email queries to other teams.

```
Section: Risks and needs
  Task: Add health needs
    Pages:  - Guidance
            - [Substance misuse]
            - [Physical health]
            - [...]
```

### Task list with new "Risks and needs" section and "Add health needs" task
risks_and_needs_section

### Health needs guidance page

Without the copy-able "Get in touch" section and email templates

health_needs_guidance

### DIFF
--- a/integration_tests/pages/apply/taskListPage.ts
+++ b/integration_tests/pages/apply/taskListPage.ts
@@ -33,4 +33,8 @@ export default class TaskListPage extends Page {
       cy.get('.app-task-list__task-name').contains(taskTitle)
     })
   }
+
+  visitTask = (taskTitle: string): void => {
+    cy.get('.app-task-list__task-name').contains(taskTitle).click()
+  }
 }
--- a/server/form-pages/apply/risks-and-needs/health-needs/guidance.ts
+++ b/server/form-pages/apply/risks-and-needs/health-needs/guidance.ts
@@ -1,0 +1,43 @@
+import type { TaskListErrors } from '@approved-premises/ui'
+import { Cas2Application as Application } from '@approved-premises/api'
+import { Page } from '../../../utils/decorators'
+import TaskListPage from '../../../taskListPage'
+
+type GuidanceBody = Record<string, never>
+
+@Page({
+  name: 'health-needs-guidance',
+  bodyProperties: [],
+})
+export default class Guidance implements TaskListPage {
+  title = `Request health information for ${this.application.person.name}`
+
+  body: GuidanceBody
+
+  constructor(
+    body: Partial<GuidanceBody>,
+    private readonly application: Application,
+  ) {
+    this.body = body as GuidanceBody
+  }
+
+  previous() {
+    return 'taskList'
+  }
+
+  next() {
+    return ''
+  }
+
+  errors() {
+    const errors: TaskListErrors<this> = {}
+
+    return errors
+  }
+
+  response() {
+    const response = {}
+
+    return response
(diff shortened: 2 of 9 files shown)


edavey:
[Health needs guidance page](https://trello.com/c/rmEOYpyy/328-health-needs-guidance-page)

integration_tests/pages/apply/taskListPage.ts
```diff
@@ -33,4 +33,8 @@ export default class TaskListPage extends Page {
       cy.get('.app-task-list__task-name').contains(taskTitle)
     })
   }
+
+  visitTask = (taskTitle: string): void => {
```
Harriethw, line 37: nice!

server/form-pages/apply/risks-and-needs/health-needs/guidance.ts
```diff
@@ -0,0 +1,43 @@
+import type { TaskListErrors } from '@approved-premises/ui'
+import { Cas2Application as Application } from '@approved-premises/api'
+import { Page } from '../../../utils/decorators'
+import TaskListPage from '../../../taskListPage'
+
+type GuidanceBody = Record<string, never>
```
Harriethw, line 6: oh cool, I haven't come across this type before - is it basically a way of saying that the `body` will be an empty object?

server/views/applications/pages/health-needs/guidance.njk
```diff
@@ -0,0 +1,36 @@
...
+    <li>Nelson Trust</li>
+  </ul>
+
+  <h2 class="govuk-heading-m">Getting in touch</h2>
+
+  <p class="govuk-body">
+    <strong>We strongly suggest you get in touch with the relevant health teams and officers now.</strong>
+  </p>
+
+  <p class="govuk-body">
```
Harriethw, line 28: minor question but I'm wondering if we should not include this paragraph and the below one? The user doesn't 'need' them and on the off chance we don't get around to doing the story to insert the copy/paste text then we would have to come back and remove these paragraphs...

Harriethw (approved):
thanks @edavey - good to see how this type of page is built.

server/form-pages/apply/risks-and-needs/health-needs/guidance.ts
edavey: yes: I started off by trying to define the type with:

```ts
type GuidanceBody = {}
```

EsLint then told me:

```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.
```

server/views/applications/pages/health-needs/guidance.njk
edavey: yes, makes sense  -- certainly it's neater.

PR #131, opened 2023-08-07
Ethnic background options

Harriethw:
This PR adds four the remaining ethnic background options:
Screenshot 2023-08-04 at 16 29 14

## notes
* hopefully not too big a PR as a lot of the changes are the same!
* I kind of ploughed through these as quickly as possible - there is a lot of repetitive code but I'm not sure it would be much help trying to build more abstractions to help us versus just copy/pasting. Interested to hear any thoughts though!
* I'm not going to open an e2e test for these different options, as a we have a happy path with 'white background' as an answer, but also open to thoughts

Screenshot 2023-08-07 at 13 45 44
Screenshot 2023-08-07 at 13 46 09
Screenshot 2023-08-07 at 13 46 16
Screenshot 2023-08-07 at 13 46 27


edavey (approved):
👍 epic in scale but as you say each page is very similar. 

I agree that it's good to stick with this verbose and repetitive implementation for now -- rather than building some kind of "ethnic background specialisation" abstraction -- as we don't know how this part of the questionnaire might change once people start using it...

PR #153, opened 2023-08-16
add intial PR template

Harriethw:
A PR to add a default template for our PRs. 

* I wanted to make sure that we don't forget to add `auditEvent`s when we add a new route - let me know if any more docs are required for that.
* Copied mostly from CAS1 but with some changes for us because we don't have preprod/prod environment.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,37 @@
+# Context
+
+<!-- Is there a Trello ticket you can link to? -->
+<!-- Do you need to add any environment variables? -->
+<!-- Is an ADR required? An ADR should be added if this PR introduces a change to the architecture. -->
+
+# Changes in this PR
+
+## Screenshots of UI changes
+
+### Before
+
+### After
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+confirm:
+
+## Pre merge checklist
+
+- [ ] Are any changes required to the e2e tests?
+- [ ] If you've added a new route, have you added a new
+  `auditEvent`? (see `server/routes/apply.ts` for examples)
+- [ ] Have all changes to any dependencies been deployed in a backwards
+    compatible way? (This includes our API, infrastructure, third-party
+    integrations and any secrets or environment variables)
```
edavey, line 28: this feels a bit broad -- approaching "please confirm that you haven't forgotten anything"!

Perhaps it includes more direct questions which we could break out as separate items? e.g.

- are there environment variables or other infrastructure configuration which needs to be included in this release?
- are there any data migrations required. Automatic or manual?
- does this rely on changes being deployed to the CAS API?

PR #155, opened 2023-08-17
Mark application as submitted

Harriethw:
API PR here: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/895 

We would like users to be able to mark applications as submitted - this is a first pass at the very minimal functionality to allow users to do this. Does not include validating the application is complete, error handling or correct design for the confirmation page. 

Screenshot 2023-08-18 at 11 49 05

Screenshot 2023-08-18 at 11 49 18

### DIFF
--- a/integration_tests/tests/apply/submit.cy.ts
+++ b/integration_tests/tests/apply/submit.cy.ts
@@ -1,0 +1,46 @@
+//  Feature: Referrer submits an application
+//
+//  Scenario: submit an application
+//    Given I am logged in
+//    And I have created an appliation
+//    And I am on the Task List
+//    When I click submit
+//    Then I should see a confirmation page
+
+import ApplicationSubmittedPage from '../../pages/apply/applicationSubmittedPage'
+import Page from '../../pages/page'
+import TaskListPage from '../../pages/apply/taskListPage'
+import { applicationFactory } from '../../../server/testutils/factories'
+
+context('Applications dashboard', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubAuthUser')
+
+    // I have created an application
+    cy.fixture('applicationData.json').then(applicationData => {
+      const application = applicationFactory.build({ id: 'abc123', data: applicationData, status: 'inProgress' })
+      cy.wrap(application).as('application')
+    })
+
+    // Given I am logged in
+    cy.signIn()
+  })
+
+  //  Scenario: submit an application
+  // ----------------------------------------------
+  it('shows the dashboard', function test() {
+    cy.task('stubApplicationGet', { application: this.application })
+    cy.task('stubApplicationSubmit', { application: this.application })
+
+    // I visit the Task List Page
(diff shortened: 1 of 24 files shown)


server/controllers/apply/applicationsController.ts
```diff
@@ -83,4 +84,24 @@ export default class ApplicationsController {
       })
     }
   }
+
+  submit(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const application = await this.applicationService.findApplication(req.user.token, req.params.id)
+      application.document = getResponses(application)
+
+      // TODO: validate that the user has confirmed information is complete
```
Harriethw, line 93: I'm not a big fan of leaving TODO's in but in this case we definitely need to come back to this 🙈

server/views/applications/taskList.njk
```diff
@@ -20,32 +20,43 @@
...
+                preventDoubleClick: true
+              })
+            }}
+          </li>
+        </ol>
+      </form>
     </div>
   </div>
-{% endblock %}
+{% endblock %}
```
patrickjfl, line 62: no newline here

server/views/applications/confirm.njk
```diff
@@ -0,0 +1,25 @@
...
+            }) }}
+
+            {{ govukButton({
+                text: "Back to dashboard",
+                href: paths.applications.index()
+                })
+            }}
+        </div>
+    </div>
+{% endblock %}
```
patrickjfl, line 25: same here

patrickjfl: (annoying nunjucks formatting issues, it generally removes them for me too 👎 )

integration_tests/tests/apply/submit.cy.ts
```diff
@@ -0,0 +1,44 @@
...
+    cy.task('stubApplicationSubmit', { application: this.application })
+
+    // I visit the Task List Page
+    const page = TaskListPage.visit(this.application)
+
+    // I click submit
+    page.clickSubmit()
+
+    // Then I see a confirmation page
+    cy.get('.govuk-panel__title').should('contain', 'Application complete')
```
patrickjfl, line 42: Is it worth creating the confirmation page and then using the Page.verifyOnPage() method here, so we keep the integration tests clear and high level?

server/utils/applications/forPagesInTask.ts
```diff
@@ -0,0 +1,27 @@
+import { FormArtifact, UiTask } from '../../@types/ui'
+import { getPage } from './getPage'
+import TaskListPage from '../../form-pages/taskListPage'
+
+export const forPagesInTask = (
```
patrickjfl, line 5: could we give this function a clearer name? I'm finding it quite confusing - the functionality is complex and the name isn't guiding me :)

patrickjfl (approved):
Looks good! Just a few minor comments 👍

integration_tests/tests/apply/submit.cy.ts
Harriethw: good spot, was being lazy :)

PR #163, opened 2023-08-22
Health needs guidance page - getting in touch section

patrickjfl:
This PR allows users to copy email templates from the health needs guidance page.

The functionality has been implemented via an extra script in the guidance template, and tested in our integration test suite.

## Before

Screenshot 2023-08-22 at 11 51 49

## After

Screenshot 2023-08-22 at 11 52 10

Screenshot 2023-08-22 at 11 52 30

Screenshot 2023-08-22 at 11 52 44

### DIFF
--- a/assets/scss/components/_copy-text.scss
+++ b/assets/scss/components/_copy-text.scss
@@ -1,0 +1,22 @@
+.copy-text {
+  > div {
+    border-left: none;
+    padding-left: 0;
+  }
+
+  &__container {
+    border: 1px solid govuk-colour('dark-grey');
+    padding: 1.25rem;
+  }
+
+  &__button {
+    color: white;
+    background: govuk-colour('blue');
+    box-shadow: 0 2px 0 govuk-colour('dark-blue');
+    margin-bottom: 0;
+
+    &:hover {
+      background: govuk-colour('dark-blue');
+    }
+  }
+}
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -23,6 +23,10 @@ export default abstract class Page {
     cy.get('button').click()
   }
 
+  clickContinue(): void {
+    cy.get('a').contains('Continue').click()
+  }
+
   shouldShowErrorMessagesForFields(fields: Array<string>): void {
     fields.forEach(field => {
       const errorMessagesLookup = errorLookups[field].empty
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -2,3 +2,11 @@ Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
   cy.request('/')
   return cy.task('getSignInUrl').then((url: string) => cy.visit(url, options))
 })
+
+Cypress.Commands.add('assertValueCopiedToClipboardContains', value => {
+  cy.window().then(win => {
+    win.navigator.clipboard.readText().then(text => {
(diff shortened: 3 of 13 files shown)


integration_tests/support/commands.ts
```diff
@@ -2,3 +2,11 @@ Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
   cy.request('/')
   return cy.task('getSignInUrl').then((url: string) => cy.visit(url, options))
 })
+
+Cypress.Commands.add('assertValueCopiedToClipboardContains', value => {
```
Harriethw, line 6: cool!

server/views/partials/email-templates/health-questions.njk
```diff
@@ -1,8 +1,12 @@
...
+<div id="health-questions-template">
+  <ol>
+    <li>Do they have any physical health needs? Please describe them and the level of severity.</li>
+    <li>Are they currently receiving any medical treatment for their physical health needs? If so, please describe it.</li>
+    <li>Are they currently taking any medication for their physical health needs? If so, please provide details.</li>
+    <li>Can they climb stairs?</li>
+    <li>Can they live independently?</li>
+    <li>What support is required?</li>
+
+    <button id="health-questions-button">Copy text</button>
```
Harriethw: copy functionality seems to work great 🥳 🙌 
but because the `Copy text` button is within the same `div` the button text gets copied to. Could the button be moved to a div outside the `drug-alcohol-questions-template` one (and same for other templates)

assets/scss/components/_copy-text.scss
```diff
@@ -0,0 +1,17 @@
+.copy-text {
```
Harriethw, line 1: would it be possible to remove the `border-left` grey styling on the `govuk-details__text` component as well? It looks like the prototype doesn't include that, and would be a bit cleaner without it...

integration_tests/pages/page.ts
```diff
@@ -20,7 +20,7 @@ export default abstract class Page {
   manageDetails = (): PageElement => cy.get('[data-qa=manageDetails]')
 
   clickSubmit(): void {
-    cy.get('button').click()
+    cy.get('a').contains('Continue').click()
```
Harriethw, line 23: I think if you set `matchCase` to `false` as an option here it would pass the rest of the integration tests? https://docs.cypress.io/api/commands/contains#Arguments 

I seem to recall Ed making some changes about checking for exact matches somewhere but I can't find it 🤔

assets/scss/components/_copy-text.scss
Harriethw: would adding `border-left: none` to the `.govuk-details__text` class here override the styles from the gov uk component? not sure how the order of precedence works...

integration_tests/pages/page.ts
patrickjfl: huh, ok thanks :)

assets/scss/components/_copy-text.scss
patrickjfl: I'm not sure but it feels a bit risky to me. Would it not be better to use the html component in this case, and then add an additional class to that div to get rid of the border?

patrickjfl: actually scratch that, I didn't realise you could add classes to the nunjucks macros 👀

Harriethw (approved):

PR #171, opened 2023-08-24
Handle LAO's based on new API types

Harriethw:
[This builds on work done by CAS1/3 ](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/974)

# Context
There are certain people in the prison system who are high profile and therefore the information about them is restricted for their protection, these people are referred to as LAOs (limited access offenders). Some people may be a regular PiP for some senior staff but an LAO for other staff.

This API PR means that if a LAO is returned from the API the UI will now see only their CRN: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804. 
Previously this would have thrown a non-user friendly error.
[Trello](https://trello.com/c/IdrDyeuM/309-cannot-list-bookings-from-premises-with-lao-offenders-%F0%9F%90%9E)

As there are many places in the application where the person object is used this means we have had to make changes to all of these places. More often than not we simply use a `Person`s name which we can simply fill in 'the person' instead but in some places it is more complicated than that, so the helper function allows a custom replacement.

# Changes in this PR
- We update to the appropriate types from the API branch
- We update the factories that use the Person object to return either a RestrictedPerson or a FullPerson
- We add some utils for dealing with the two new types of person
- We adjust the People Controller to throw an error if a restricted person is returned from the API
- We call the new utility function every time we need the name of the person on the application.

## Screenshots of UI changes
Screenshot 2023-08-25 at 11 09 04


server/@types/shared/models/PersonType.ts
```diff
@@ -0,0 +1,6 @@
+/* generated using openapi-typescript-codegen -- do no edit */
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+export type PersonType = 'FullPerson' | 'RestrictedPerson' | 'UnknownPerson' | 'FullPersonInfo' | 'RestrictedPersonInfo';
```
patrickjfl, line 6: I know this is out of scope for this PR but it feels a bit weird to me that 'FullPerson' and 'FullPersonInfo' belong in the same type...

server/controllers/apply/applicationsController.ts
```diff
@@ -52,7 +53,7 @@ export default class ApplicationsController {
   }
 
   private ineligibleViewParams(application: Cas2Application): Record<string, string | Cas2Application> {
-    const panelText = `${application.person.name} is not eligible for CAS-2 accommodation`
+    const panelText = `${nameOrPlaceholderCopy(application.person)} is not eligible for CAS-2 accommodation`
```
patrickjfl, line 56: how placeholder is the text? (is it worth capitalising 'the' here?)

server/controllers/peopleController.ts
```diff
@@ -17,14 +19,19 @@ export default class PeopleController {
       if (crn) {
         try {
           const person = await this.personService.findByCrn(req.user.token, crn)
-          const application = await this.applicationService.createApplication(req.user.token, person.crn)
 
+          if (!isFullPerson(person)) {
+            throw new RestrictedPersonError(crn)
```
patrickjfl, line 24: so, in this case, do we always want to prevent the user from continuing? Or do we want to distinguish between a restricted person that the referrer has permission to refer, and a restricted person that the referred doesn't have permission to refer? I think this is cas1's approach: https://trello.com/c/IdrDyeuM/309-cannot-list-bookings-from-premises-with-lao-offenders-%F0%9F%90%9E

patrickjfl (commented):
Looks good, just a few comments. Also is it worth adding an integration test to the find_by_crn file to capture the restricted person case?

server/controllers/peopleController.ts
Harriethw: good question - we'll need to raise it with HMPPS to confirm and come back to this - for now I'm assuming we don't let the application be created if it's an LAO

server/controllers/apply/applicationsController.ts
Harriethw: hopefully the user should never see the placeholder text in CAS2 because we shouldn't have an application with an LAO...  but I think we need to account for the person becoming an LAO during an application? And because of the types now coming from API we always need to check this... 

tbh I think it the solution could be looked at again with a bit more time as I'm not 100% sure we are handling everything in the best way, but want to unblock us from being able to update API types now.

server/controllers/peopleController.ts
Harriethw: slack thread here https://mojdt.slack.com/archives/C04T6SC31HA/p1692959904971339?thread_ts=1692959679.970929&cid=C04T6SC31HA

patrickjfl: ok thanks 👍

server/controllers/apply/applicationsController.ts
patrickjfl: ok yep, makes sense

server/@types/shared/models/PersonType.ts
Harriethw: yeah I hadn't seen that... I'm not entirely sure I understand what the difference is between `FullPerson` and `FullPersonInfo` added here https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804

Harriethw: I think it might be left in from some changes that were made... `Person` was going to become `PersonInfo` but is now just `Person` (and same for other types?)

server/controllers/apply/applicationsController.ts
Harriethw: think I have now capitalised wherever the name is the first word of a sentence

patrickjfl (approved):
LGTM 👍

PR #173, opened 2023-08-25
Risk to self guidance page

patrickjfl:
This PR adds a risk to self task and a guidance page. This page will attempt to retrieve OASys data for the person, and if successful will invite the user to import that data, creating a copy that will be saved to the application. If no OASys data is found the user is informed and invited to either update OASys or complete the section manually.

## Before

N/A

## After

Screenshot 2023-08-25 at 09 15 15

Screenshot 2023-08-25 at 09 16 17

PR #176, opened 2023-08-29
Refactor Risk to Self OASys transformation 

Harriethw:
We would like to check if there is OASys information for an applicant on the Risk to Self section.
If there is data, when the user clicks 'save and continue' we populate the relevant answers in the rest of the task.

[We previously used an existing function ](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/173)on the CAS API which gets several sections from the report, and does not include one required part of the Risk to Self section (historical risk of self-harm). 

This PR:
* Uses work from this API PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/920 to retrieve only the relevant section of OASys
* adds a skeleton 'Vulnerability' page so that we can see answers populating.

Screenshot 2023-08-29 at 14 11 48

Screenshot 2023-08-29 at 14 11 07

### DIFF
--- a/server/form-pages/apply/risks-and-needs/risk-to-self/riskToSelfGuidance.ts
+++ b/server/form-pages/apply/risks-and-needs/risk-to-self/riskToSelfGuidance.ts
@@ -4,6 +4,7 @@ import { Page } from '../../../utils/decorators'
 import TaskListPage from '../../../taskListPage'
 import { DateFormats } from '../../../../utils/dateUtils'
 import { nameOrPlaceholderCopy } from '../../../../utils/utils'
+import Vulnerability from './vulnerability'
 
 type GuidanceBody = Record<string, never>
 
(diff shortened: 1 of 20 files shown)


Harriethw:
https://trello.com/c/rT4kUDt2/470-rtos-tech-task-risk-to-self-skeleton

Harriethw:
https://trello.com/c/mzr3lwWs/481-rtos-oasys-guidance-page-mvp

server/form-pages/apply/risks-and-needs/risk-to-self/riskToSelfGuidance.ts
```diff
@@ -68,18 +69,21 @@ export default class RiskToSelfGuidance implements TaskListPage {
...
-    try {
-      oasysSections = await dataServices.personService.getOasysSections(token, application.person.crn)
-
-      taskDataJson = JSON.stringify(RiskToSelfGuidance.getTaskData(oasysSections))
-    } catch (e) {
-      if (e.status === 404) {
-        oasysSections = null
-      } else {
-        throw e
+    if (!application.data['risk-to-self'] || Boolean(Object.keys(application.data['risk-to-self']).length < 1)) {
```
patrickjfl, line 76: just for my own confusion, what case is the second condition covering here?

patrickjfl (approved):
Nice work 🙂

server/form-pages/apply/risks-and-needs/risk-to-self/riskToSelfGuidance.ts
Harriethw: ah good spot, it doesn't need to be in there, it was just an issue with how integration tests were setup (setting the `risk-to-self` data as empty object, when in real life it wouldn't exist at all)

PR #185, opened 2023-09-01
Add ACCTs to Risk to Self

Harriethw:
# Context

[Trello card](https://trello.com/c/sa4y9S6w/505-rtos-mvp-manual-acct)

Referrers need to be able to add to a list of ACCTs (Assessment, Care in Custody and Teamwork). Our current implementation will allow them to do this manually. 
* From the 'ACCT' page, they can open a form to add a new ACCT. 
* Once that form is submitted that data is saved and they are taken back to the 'ACCT' page to see it rendered. 
* 'save and continue' on the ACCT page itself does not need to save anything but just continue to next page

# Changes in this PR

**ACCT Page**
Screenshot 2023-09-01 at 15 45 25

**ACCT Data page** 
Screenshot 2023-09-01 at 15 46 16

This is another case (like `risk-to-self/guidance.njk`) where we have to deviate from the regular flow of the "Task List system". When normal form pages are saved they will PUT the data from fields on the form and the controller will use it to update the data for that particular page in the application, overwriting the previous data saved. 

In this case:
* we need to change the regular data on a form to be an array instead of an object
* and we need to preserve previously saved data

The `riskToSelfGuidance` page will always be the first ever page in a Task and it needs to update other pages - to do this it collates all the Task data as a hidden field on the form and PUTs it to the application (which is does when the form is submitted):  

```
<form action="{{ paths.applications.update({ id: applicationId }) }}?_method=PUT" method="post">
          <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
          <input type="hidden" name="pageName" value="{{ page.name }}"/>
          <input type="hidden" name="taskData" value="{{ page.taskData }}"/>
          <input type="hidden" name="taskName" value="{{ page.taskName }}"/>
...
```

This ACCT data however could come at a point when there is already data against the Task. It also doesn't need to update any other pages in the Task, only it's own data.

I'm therefore suggesting that the `acctData` page is responsible for storing and updating all of the ACCTs added to the application, and the `acct` page is responsible for rendering the data. This means we can use the regular Task list system to update the data for that page, but setting all the ACCTs as hidden inputs on the `acctData` form and using bracket notation to add the data to a list:

```
{% for acct in page.existingAccts %}
    {% set acctIndex = loop.index %}

    {% for date in acct.createdDate %}
      <input type="hidden" name="acctData[{{ acctIndex }}][createdDate][]" value="{{ date }}"/>
    {% endfor %}

    {% for date in acct.expiryDate %}
      <input type="hidden" name="acctData[{{ acctIndex }}][expiryDate][]" value="{{ date }}"/>
    {% endfor %}

    <input type="hidden" name="acctData[{{ acctIndex }}][referringInstitution]" value="{{ acct.referringInstitution }}"/>

    <input type="hidden" name="acctData[{{ acctIndex }}][acctDetails]" value="{{ acct.acctDetails }}"/>
 {% endfor %}
```

When the `acctData` page is submitted, it will update the application data to look like this:
```
"risk-to-self: {
    "acct-data": {
      "acctData": [
        {
          "createdDate": ["10", "10", "2013"],
          "expiryDate": ["1", "2", "2012"],
          "referringInstitution": "HMPPS prison",
          "acctDetails": "ACCT details"
        }
      ]
    }
}
```

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -124,10 +124,11 @@
     "historical-risk": {
       "historicalRiskDetail": "example answer historical risk "
     },
-    "acct-data": [
-      {
-        "acctDetails": "some answer"
-      }
-    ]
+    "acct-data": {
+      "acctData": [
+        { "acctDetails": "222", "expiryDate": ["1", "2", "2012"] },
+        { "acctDetails": "111", "expiryDate": ["1", "12", "2012"] }
+      ]
+    }
   }
 }
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -126,8 +126,12 @@
     },
     "acct-data": {
       "acctData": [
-        { "acctDetails": "222", "expiryDate": ["1", "2", "2012"] },
-        { "acctDetails": "111", "expiryDate": ["1", "12", "2012"] }
+        {
+          "createdDate": ["10", "10", "2013"],
+          "expiryDate": ["1", "2", "2012"],
+          "referringInstitution": "HMPPS prison",
+          "acctDetails": "ACCT details"
+        }
       ]
     }
   }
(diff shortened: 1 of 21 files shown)


Harriethw:
closing to use a different implementation

PR #186, opened 2023-09-04
413 acct custom form

Harriethw:
# Context

# Changes in this PR

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


server/utils/dateUtils.test.ts
```diff
@@ -158,4 +158,60 @@ describe('DateFormats', () => {
       expect(differenceInDays).toHaveBeenCalledWith(date1, date2)
     })
   })
+
+  describe('dateAndTimeInputsToUiDate', () => {
+    it('converts a date and time input object to a human readable date', () => {
+      const dateTimeInputs = { 'key-day': '11', 'key-month': '11', 'key-year': '2022' }
+
+      expect(DateFormats.dateAndTimeInputsToUiDate(dateTimeInputs, 'key')).toEqual('11/11/2022')
```
edavey, line 166: probably I'm not understanding, but would it be clearer if day and month were not both `'11'` here?

server/services/applicationService.ts
```diff
@@ -76,6 +76,35 @@ export default class ApplicationService {
...
+      const pageName = getPageName(page.constructor)
+      const taskName = getTaskName(page.constructor)
+
+      /* eslint-disable no-underscore-dangle */
+      delete request.body._csrf
+      delete request.body.pageName
+      delete request.body.taskName
+
+      application.data = application.data || {}
+      if (application.data[taskName] && application.data[taskName][pageName]) {
```
edavey, line 97: For people coming fresh to this, do you think it might be clearer if these tests were wrapped in functions with descriptive names?

e.g. I _think_ that `(application.data[taskName] && application.data[taskName][pageName]) ` is asking `isAppendingAnItem()`?

server/controllers/apply/applicationsController.test.ts
```diff
@@ -218,6 +218,82 @@ describe('applicationsController', () => {
...
+
+        console.log('request is', request)
+
+        await requestHandler({ ...request }, response)
+
+        expect(applicationService.appendToList).toHaveBeenCalledWith(page, request)
+
+        expect(response.redirect).toHaveBeenCalledWith(
+          paths.applications.pages.show({ id: request.params.id, task: 'example-task', page: 'next-page' }),
+        )
```
edavey: really easy to follow, thks!

server/services/applicationService.test.ts
```diff
@@ -155,6 +155,84 @@ describe('ApplicationService', () => {
...
+        await service.appendToList(page, requestWithBody)
+
+        expect(getApplicationUpdateData).toHaveBeenCalledWith({
+          ...application,
+          data: {
+            'some-task': {
+              'some-page': [{ question: 'answer' }, { newQuestion: 'new answer' }],
+            },
+          },
+        })
```
edavey, line 215: very clear!

server/form-pages/apply/risks-and-needs/risk-to-self/acctData.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    })
+
+    const requiredFields = [
+      ['createdDate', 'Add a valid created date, for example 2 3 2013'],
+      ['expiryDate', 'Add a valid expiry date, for example 2 3 2013'],
+      ['referringInstitution', 'Add a referring institution'],
+      ['acctDetails', 'Enter the details of the ACCT'],
+    ]
+
+    it.each(requiredFields)('it includes a validation error for %s', (field, message) => {
```
edavey, line 69: neat!

integration_tests/tests/apply/risks_and_needs/risk-to-self/acct.cy.ts
```diff
@@ -30,14 +34,22 @@ context('Visit "Risks and needs" section', () => {
     cy.task('stubAuthUser')
 
     cy.fixture('applicationData.json').then(applicationData => {
-      applicationData['risk-to-self'] = {}
+      delete applicationData['risk-to-self']
```
edavey, line 37: this is a handy way to clear the fixture data

server/views/applications/pages/risk-to-self/acct.njk
```diff
@@ -1,6 +1,57 @@
...
+            html: acct.acctDetails
+          }
+        }
+      ]
+    }) }}
+
+  {% endfor %}
+
+  {{ govukButton({
+    text: "Select ACCT notes",
```
edavey, line 52: I think this should probably be something like "Add ACCT note"?

edavey (approved):
Great work, so clear, lovely unit tests!

I've left a few comments / suggestions...

edavey:
### Two types of form submission

I struggled initially to grasp the distinction here between the two types of form submission:

**1. Regular 'page' submission to the pages controller**

```ts
paths.applications.pages.update({
  id: applicationId, 
  task: task, 
  page: pageName
})
```

e.g.

`PUT /applications/{applicationId}/tasks/confirm-eligibility/pages/confirm-eligibility`

In this case the task and page information is in the path.


**2. Special cases of submission direct to the applications controller**

```ts
paths.applications.update({ 
  id: applicationId 
})
```

e.g. 

`PUT /applications/{applicationId}`

In this case the task and page data is supplied in hidden form fields.

We submit form data direct to the application controller in more complex scenarios e.g.

- when we are importing data in bulk, behind the scenes, from upstream services like OASys
- when we need to build up an arbitrarily long list of items, as in this case where the applicant can add one ACCT record after the other.

edavey:
### Naming of "guidance" page
Would "import" be clearer? I was initially struggling to understand why a _guidance_ page (which perhaps sounds like pure content) would need special behaviour...

server/services/applicationService.test.ts
patrickjfl, line 215: not really related to this PR, but if we're sending the whole application to the API every time we save, is it really a PUT request?

patrickjfl (approved):
Makes sense to me 👍

server/services/applicationService.test.ts
edavey: which method were you thinking might be more appropriate?

In general, I would say that PUT is appropriate for any form of update. 

However, some teams / frameworks prefer to use PATCH for a partial update as distinct from a full replacement, which I think is technically more correct. Perhaps we should use a PATCH here? But I would say that change could follow on as an enhancement.

A couple of references:

- [PUT vs PATCH & PUT vs POST](https://dev.to/mehmehmehlol/put-vs-patch-put-vs-post-56i9)
- [Using PATCH in express](https://www.tabnine.com/code/javascript/functions/express/Express/patch)

doc/architecture/decisions/0002_custom_forms.md
```diff
@@ -0,0 +1,63 @@
...
+## Decision
+
+We will submit form data direct to the application controller in more complex
+scenarios e.g.
+
+- when we are importing data in bulk, behind the scenes, from upstream services like OASys
+- when we need to build up an arbitrarily long list of items, as in this case
+  where the applicant can add one ACCT record after the other.
+
+Where relevant we will put these forms in their own folder, e.g. a `custom-data`
```
patrickjfl, line 61: I like this idea, it's very clear signposting, but could we put these in a `custom-data` data directory within the task directory they belong in? Then we can preserve both categories (task, and form type) in organising these files.

patrickjfl (approved):
Ok with this but I think it would be better to keep these files within the task the belong to (albeit still in a `custom-data` directory).

doc/architecture/decisions/0002_custom_forms.md
edavey: If I understand correctly, we suggest `custom-data` as the directory name in the ADR but use `custom-forms` in the commit?

Harriethw: > If I understand correctly, we suggest custom-data as the directory name in the ADR but use custom-forms in the commit?

whoops! changed to `custom-forms`
I also want to rename the `acct-data` page to something more like `add-acct` but figured this PR was big enough already and can do it in a subsequent one :)

server/services/applicationService.test.ts
patrickjfl: I think it just feels weird to me because we're passing around the whole JSON blob, so in a sense it feels more like a POST. But I guess that's just how it is when you have this data type? Whatever the operation (unless it's DELETE), you run your filtering logic in the UI, and then send the whole thing to the API...?

Harriethw:
Thank you so much @edavey and @patrickjfl for your feedback! 

I think we almost ready to go, there is just some kind of bug with the ability to submit the application because of the `acct-data` page 🤔  I think it's to do with the page being an array rather than object when we transform the data on submit...

doc/architecture/decisions/0002_custom_forms.md
patrickjfl: Ah I think I misread the file directory system and you were already doing what I suggested 🙂

PR #202, opened 2023-09-08
Remove applicant's name from Document title element

Harriethw:
# Context

https://trello.com/c/PcJDneZY/516-hide-pii-in-page-titles

We currently use a form page's `title` field to populate the document
title `<title>` element.

In order to protect PII leaking through Google Analytics (and
potentially other platforms) we need to remove the person's name from
this element. We still want to preserve the name in the page heading as
it provides a better user experience.

I'm proposing introducing a new required field on a `TaskListPage` that will
account for what the document title should be, so that we can provide a
custom one without the person's name.

The `<title>` is then set on the shared page `layout` template.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


server/form-pages/taskListPage.ts
```diff
@@ -13,6 +13,8 @@ export interface TaskListPageInterface {
 }
 
 export default abstract class TaskListPage {
+  abstract documentTitle: string
```
Harriethw, line 16: I'm thinking this could be better named to show what it intends... `titleWithoutPersonalInformation` ??

libuk:
I haven’t put together a PR yet, but I think it’s possible to rename the page (document) titles and leave the form page titles unchanged by using something like the following in the `layout.njk` template all other form pages inherit from.


```
{% set applicationPageTitle = applicationName + " - " + page.title %}
{% set pageTitle = pageTitle | default(applicationPageTitle) %}
```

Then overwrite an individual page:

```
{% set pageTitle = "This person" %}
```

Harriethw:
@libuk so have the document title as an optional field essentially? And on each template where we wanted to override it we'd have to add a `set`? 
I'm wondering as 90% of our pages will likely need to set it, if it's safer to have it as required, and if it's no different to the `page.title` then it can just be set as that in the typescript file.

Harriethw:
also just thinking about the cognitive overload on the devs having to make 100s of these pages, personally I would prefer to set in the typescript file and have clear requirements for what's needed.

libuk:
> I'm wondering as 90% of our pages will likely need to set it, if it's safer to have it as required, and if it's no different to the page.title then it can just be set as that in the typescript file.

@Harriethw Yeah, it's a good point. I'm happy to go with that approach.

server/form-pages/apply/risks-and-needs/risk-of-serious-harm/additionalRiskInformation.ts
```diff
@@ -11,6 +11,8 @@ type AdditionalRiskInformationBody = Record<string, never>
   bodyProperties: [],
 })
 export default class AdditionalRiskInformation implements TaskListPage {
+  documentTitle = 'Additional risk information for the person'
```
patrickjfl, line 14: It seems like almost all of the example titles sub out the person's name for 'the person' - could we make a util function to generate both of them at the same time?

patrickjfl (approved):
LGTM 👍

server/form-pages/apply/risks-and-needs/risk-of-serious-harm/additionalRiskInformation.ts
Harriethw: I tooled around with a utility function, or something like
``` 
documentTitle = "Import the person's risk to self data from OASys"

title = this.documentTitle.replace('the person', this.personName)
```
but everything I tried just seemed over-complicated or less readable than just writing the string.. would be interested in any suggestions

patrickjfl: I quite like the utility function idea because to me it flags that we are deliberately setting these values differently, but I see what you're saying. It could be something like this:

```
function generateDocumentAndPageTitles(documentTitle: string, name: string) {
  const pageTitle = documentTitle.replace('the person', this.personName)

  return { documentTitle, pageTitle }
}
```

Harriethw: ooh I also figured out how to add documentation to a field in typescript so that will add some immediate context so future devs won't have to look in git commits

PR #205, opened 2023-09-11
507 remove acct

Harriethw:
# Context

# Changes in this PR

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


server/paths/apply.ts
```diff
@@ -21,6 +21,7 @@ const paths = {
     },
     update: singleApplicationPath,
     appendToList: singleApplicationPath.path('list'),
+    removeFromList: singleApplicationPath.path('list').path('remove'),
```
Harriethw, line 24: I'm not sure what the endpoint should be, this doesn't feel right but couldn't think of what else 🤔

server/routes/apply.ts
```diff
@@ -23,6 +23,9 @@ export default function applyRoutes(controllers: Controllers, router: Router, se
   put(paths.applications.appendToList.pattern, applicationsController.appendToList(), {
     auditEvent: 'UPDATE_APPLICATION_LIST',
   })
+  get(paths.applications.removeFromList.pattern, applicationsController.removeFromList(), {
```
patrickjfl, line 26: should this not be a PUT?

server/paths/apply.ts
patrickjfl: maybe would be clearer to make this ACCT specific? e.g. `singleApplicationPath.path('acct-notes').path('remove')`

patrickjfl (approved):
Nice and clear 👍

server/routes/apply.ts
edavey: yes I think PUT or PATCH is more appropriate for this request which changes a list of items...

server/views/applications/pages/risk-to-self/acct.njk
```diff
@@ -10,43 +10,58 @@
...
+            text: acct.referringInstitution
           },
-          value: {
-            html: acct.createdDate
-          }
+        actions: {
+          items: [
+            {
+              href: acct.removeLink,
+              text: "Remove",
```
edavey, line 26: If you were going to make the removal request a PUT request then I'm guessing you'd pass in a form here to the `actions` property somehow?

integration_tests/pages/apply/risks-and-needs/risk-to-self/acctPage.ts
```diff
@@ -23,4 +23,12 @@ export default class AcctPage extends ApplyPage {
     cy.get('.govuk-summary-list').contains('10/10/2013')
     cy.get('.govuk-summary-list').contains('ACCT details')
   }
+
+  hasNoAccts(): void {
+    cy.get('.govuk-body').contains('No ACCT notes have been added.')
```
edavey, line 28: might it also be worth verifying that there are no list items present?

edavey (approved):
this looks great! 

I do think the removal request ought to be a PUT or a PATCH though. Do you think it would be "easy" to achieve?

Harriethw:
@patrickjfl @edavey thank you for the notes! 

I agree I would really like it to be a PUT, but it seems overly complicated to make that happen... and I've pushed a wip commit which shows a bit of what we'd have to do. 
1) we'd have to put a form element as you've said in the `action` for the card. This should be simple, except that it would be nested within the page layout's `form` element which means it's button will take the default action for the parent form
2) ok, so we could overwrite the page layout template `content` block so that we don't use the parent `form` element... but because the template is nested within the `risk-to-self.njk` template, which itself is nested within `layout.njk`, to override the `content` block we need to copy over the whole `risk-to-self` template, which I don't really want to do. 

Just seems like a lot of changes for the sake of a clearer HTTP request type...but interested to hear if there's a simple solution I'm missing!

server/paths/apply.ts
patrickjfl: does it not still make sense to differentiate paths for different add/remove to list tasks? E.g. so we can pick up more useful and granular data from audit events...?

patrickjfl:
> @patrickjfl @edavey thank you for the notes!
> 
> I agree I would really like it to be a PUT, but it seems overly complicated to make that happen... and I've pushed a wip commit which shows a bit of what we'd have to do.
> 
>     1. we'd have to put a form element as you've said in the `action` for the card. This should be simple, except that it would be nested within the page layout's `form` element which means it's button will take the default action for the parent form
> 
>     2. ok, so we could overwrite the page layout template `content` block so that we don't use the parent `form` element... but because the template is nested within the `risk-to-self.njk` template, which itself is nested within `layout.njk`, to override the `content` block we need to copy over the whole `risk-to-self` template, which I don't really want to do.
> 
> 
> Just seems like a lot of changes for the sake of a clearer HTTP request type...but interested to hear if there's a simple solution I'm missing!

Hmmmm agree it's a big headache, I hadn't considered all the implications. One thought I have is, how set is the design for this page? Is it GDS-y enough to just allow users to click to remove these notes without any kind of confirmation screen? Because if we decide that we do need a confirmation screen, then these problems would go away...

Harriethw:
> Is it GDS-y enough to just allow users to click to remove these notes without any kind of confirmation screen? 

This is a [standard GDS Summary Card](https://design-system.service.gov.uk/components/summary-list/card-with-actions/index.html). There is guidance that says to consider adding a warning if `the action cannot easily be undone or might have serious consequences' but I don't think this warrants it - the worst consequence would be that they have to spend a couple of minutes filling in the data again.

server/paths/apply.ts
Harriethw: hmm yeah that's interesting point and has made me realise we aren't sending over the query param data to the audit service for these application methods (`appendToList` `removeFromList` and `update`) as I thought we were...

Will have a look at some options

patrickjfl:
> 

thanks, that makes sense. Headaches it is then. I have a vague memory something this coming up on CAS3 at some point and IIRC Bryan went with a similar solution.

Harriethw:
> yes, here it is: https://mojdt.slack.com/archives/C03K0HB0LBE/p1679498864790919

good to know there's a bit of a precedent! I think it happens a lot with nunjucks/express. 
So - let's leave the 'remove' link as a regular `get` request/hyperlink as it's less of a headache/complex than making a lot of brittle changes to the html templates. In the end it's still a `put` request to the API. 

We will change the `paths` for the `appendToList` and `removeFromList` to have a more specific structure, including the page and task name in params.

PR #217, opened 2023-09-19
Add rosh risk summary

patrickjfl:
# Changes in this PR

Updates the RoSH summary page to render a RoSH summary card, based on data that is retrieved from the RoSH OASys import page.

NB following the prototype, I have allowed users to add additional comments to an 'unknown RoSH' summary, although this feels a bit strange to me.

## Screenshots of UI changes

Screenshot 2023-09-19 at 09 39 08

Screenshot 2023-09-19 at 09 40 17

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/pages/apply/risks-and-needs/risk-of-serious-harm/roshSummaryPage.ts
+++ b/integration_tests/pages/apply/risks-and-needs/risk-of-serious-harm/roshSummaryPage.ts
@@ -2,6 +2,8 @@ import { Cas2Application as Application } from '../../../../../server/@types/sha
 import ApplyPage from '../../applyPage'
 import { nameOrPlaceholderCopy } from '../../../../../server/utils/utils'
 import paths from '../../../../../server/paths/apply'
+import { RoshRisks } from '../../../../../server/@types/shared/models/RoshRisks'
+import { DateFormats } from '../../../../../server/utils/dateUtils'
 
 export default class RoshSummaryPage extends ApplyPage {
   constructor(private readonly application: Application) {
@@ -21,5 +23,33 @@ export default class RoshSummaryPage extends ApplyPage {
         page: 'summary',
       }),
     )
+  }
+
+  shouldShowRiskData = (risks: RoshRisks & { dateOfOasysImport: string }): void => {
+    cy.get('p').contains(
+      `Imported from OASys on ${DateFormats.isoDateToUIDate(risks.dateOfOasysImport, {
+        format: 'medium',
+      })}, last updated on ${DateFormats.isoDateToUIDate(risks.lastUpdated, {
+        format: 'medium',
+      })}`,
+    )
+
+    cy.get('h3').contains(`${risks.overallRisk.toLocaleUpperCase()} RoSH`)
+
+    cy.get('.rosh-widget__table').within($row => {
+      cy.wrap($row).get('th').contains('Children').get('td').contains(risks.riskToChildren, { matchCase: false })
+      cy.wrap($row).get('th').contains('Public').get('td').contains(risks.riskToPublic, { matchCase: false })
+      cy.wrap($row).get('th').contains('Known adult').get('td').contains(risks.riskToKnownAdult, { matchCase: false })
+      cy.wrap($row).get('th').contains('Staff').get('td').contains(risks.riskToStaff, { matchCase: false })
+    })
+  }
+
+  shouldShowUnknownRoshCard = (): void => {
+    cy.get('h3').contains('UNKNOWN LEVEL RoSH')
+    cy.get('p').contains('Something went wrong. We are unable to show RoSH information.')
+  }
+
+  clickAddComments = (): void => {
+    cy.get('span').contains('Add comments to this summary').click()
   }
 }
--- a/integration_tests/tests/apply/risks_and_needs/risk-of-serious-harm/rosh_summary.cy.ts
+++ b/integration_tests/tests/apply/risks_and_needs/risk-of-serious-harm/rosh_summary.cy.ts
@@ -11,8 +11,17 @@
 //  Scenario: view 'RoSH summary' page
 //    Then I see the "RoSH summary" page
 //
+//  Scenario: view risk summary data
+//    Given there is risk summary data
+//    Then I see the data presented on the page
+//
+//  Scenario: view 'unknown RoSH' card
+//    Given there is no risk summary data
+//    Then I see the unknown RoSH card
+//
 //  Scenario: navigate to next page in "Risk of serious harm" task
-//    When I continue to the next task / page
+//    When I add an additional comment
+//    And I continue to the next task / page
 //    Then I see the "Risk to others" page
 
 import Page from '../../../../pages/page'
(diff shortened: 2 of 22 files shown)


patrickjfl:
[ROSH - RoSH risk summary from OASys](https://trello.com/c/dl2n6Ny1/343-rosh-rosh-risk-summary-from-oasys)

integration_tests/tests/apply/risks_and_needs/risk-of-serious-harm/rosh_summary.cy.ts
```diff
@@ -11,32 +11,52 @@
...
+  const risks = { ...roshRisksFactory.build(), dateOfOasysImport }
 
   beforeEach(function test() {
     cy.task('reset')
     cy.task('stubSignIn')
     cy.task('stubAuthUser')
 
     cy.fixture('applicationData.json').then(applicationData => {
-      applicationData['risk-of-serious-harm'] = {}
+      applicationData['risk-of-serious-harm'].summary = risks
```
Harriethw, line 43: I was thinking going forward to have the `applicationData.json` file as an example of a full application with as much data as possible - it will be a handy reference for us when we want to check data structures and ensure we're rendering everything in the `check your answers`. So rather than adding the data in the test it would be in the data by default, and then the test can just `delete` keys on the data to test the unhappy path. Does that make sense?

server/services/personService.ts
```diff
@@ -31,4 +32,17 @@ export default class PersonService {
...
     return rosh
   }
+
+  async getRoshRisks(token: string, crn: string): Promise<RoshRisks> {
+    const personClient = this.personClientFactory(token)
+
+    const risks = await personClient.risks(crn)
+
+    return {
+      ...risks.roshRisks?.value,
```
Harriethw, line 42: I think there's at least one unhappy path not accounted for (and not covered by unit tests on lines 42-43.)

When I try to run this locally the data I get back from the API does not have `roshRisks` and so `lastUpdated` is '', which is causing an `Invalid Date` error on the UI.

server/services/personService.ts
```diff
@@ -31,4 +32,17 @@ export default class PersonService {
...
+      lastUpdated: risks.roshRisks?.value?.lastUpdated
```
Harriethw, line 43: should the service be responsible for transforming the date here, or do we think better for it to just return the data we get back from the client? I'm wondering if makes more sense for the levels closer to the UI to decide how a date is rendered. Also feel a bit uncomfortable about setting it to an empty string when there are no `roshRisks`...

integration_tests/tests/apply/risks_and_needs/risk-of-serious-harm/rosh_summary.cy.ts
patrickjfl: yes, good idea, will change

integration_tests/pages/apply/risks-and-needs/risk-of-serious-harm/roshSummaryPage.ts
```diff
@@ -22,4 +24,32 @@ export default class RoshSummaryPage extends ApplyPage {
...
+        format: 'medium',
+      })}, last updated on ${DateFormats.isoDateToUIDate(risks.lastUpdated, {
+        format: 'medium',
+      })}`,
+    )
+
+    cy.get('h3').contains(`${risks.overallRisk.toLocaleUpperCase()} RoSH`)
+
+    cy.get('.rosh-widget__table').within($row => {
+      cy.wrap($row).get('th').contains('Children').get('td').contains(risks.riskToChildren, { matchCase: false })
```
edavey, line 40: Is this `$row` variable name using the `$` to denote the fact that it's been assigned a DOM element?

edavey (approved):
👍 just a couple of small queries

server/services/personService.ts
patrickjfl: thanks - I agree this needed changing and have done so as discussed :)

server/testutils/factories/risks.ts
```diff
@@ -28,6 +28,15 @@ const roshRisksFactory = Factory.define<PersonRisks['roshRisks']>(() => ({
   },
 }))
 
+export const roshRisksFactory = Factory.define<RoshRisks>(() => ({
+  overallRisk: faker.helpers.arrayElement(riskLevels),
+  riskToChildren: faker.helpers.arrayElement(riskLevels),
+  riskToPublic: faker.helpers.arrayElement(riskLevels),
+  riskToKnownAdult: faker.helpers.arrayElement(riskLevels),
+  riskToStaff: faker.helpers.arrayElement(riskLevels),
+  lastUpdated: DateFormats.dateObjToIsoDate(faker.date.past()),
```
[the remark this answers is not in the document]
patrickjfl: in principle yes, although really most of the file is not needed, and those commits are (mine) from three months ago...

integration_tests/pages/apply/risks-and-needs/risk-of-serious-harm/roshSummaryPage.ts
edavey: it used to be quite a common convention (possibly still is?) and I do think it's helpful. It believe it goes back to the JQuery alias of `$(selector)` and would signal that the variable had a JQuery object assigned. That history may or may not make adopting the convention attractive?

patrickjfl: Sounds helpful to me! I'm sold 🙂

server/testutils/factories/risks.ts
```diff
@@ -10,13 +10,13 @@ const riskEnvelopeStatuses: Array<RiskEnvelopeStatus> = ['retrieved', 'not_found
...
   crn: `C${faker.number.int({ min: 100000, max: 999999 })}`,
-  roshRisks: roshRisksFactory.build(),
+  roshRisks: roshRisksEnvelopeFactory.build(),
   mappa: mappaFactory.build(),
   flags: flagsFactory.build(),
   tier: tierEnvelopeFactory.build(),
 }))
 
-const roshRisksFactory = Factory.define<PersonRisks['roshRisks']>(() => ({
+export const roshRisksEnvelopeFactory = Factory.define<PersonRisks['roshRisks']>(() => ({
```
edavey, line 19: maybe the commit message here ought to be tweaked now that your have introduced a "risks envelope factory". ie. rather than

> The risk factories file here contains much more complexity than we need for our service, but because we are calling an API endpoint that returns types of risk data that we don’t need, for now we need to retain this complexity.

maybe something to the effect that you've opted to return a subset of the data from the `PersonService.getRoshRisks()` and introduce a more specialised factory?

edavey (approved):
Looks good! I've just suggested a tweak to the first commit msg

Harriethw (approved):
nice work, great to have this in the form now 💐

server/testutils/factories/risks.ts
patrickjfl: yes, although I've actually just renamed the existing `roshRisksFactory` to be more specific. The complexity comment is more referring to the file in general, which contains factories for various bits of data that are required by the application type but not by our service. But I could just remove that comment if it's confusing?

PR #228, opened 2023-09-21
Risk to self errors

Harriethw:
# Context

I noticed when filling in the Risk To Self task that the OASyS import date was disappearing from questions that I knew should have them, for example...
date is present when I first see question...
Screenshot 2023-09-22 at 11 52 31

after saving the question and returning, date is gone...

Screenshot 2023-09-22 at 11 52 37

# Changes in this PR

This PR has two parts:

## 1. changing the way OASyS import date is saved

This issue was happening because the import date was saved against each auto-populated page in
a task as a field `oasysImportDate`:

```
task-name: {
  page-name: {
     oasysData: 'some imported data'
     oasysImportDate: '2023-09-19'
  }
}
```

However, when a user would then press 'save and continue' to submit
their answer, the field was not included in the body of
the form and so overwritten when the data was updated.

```
task-name: {
  page-name: {
    oasysData: 'some updated imported data'
    confirmation: 'confirmed'
  }
}

```

I think it makes sense for the date to 'belong' to an `oasys-import` page
anyway, as it is more like metadata for the section.

So I've refactored import form pages to save it on the task, and the pages to look for the date there:

```
task-name: {
    oasys-import: {
        oasysImportDate: ''2023-09-19'
    }
    page-name: {
        oasysData: 'some updated imported data'
        confirmation: 'confirmed'
    }
}
```
Because we in theory should only show the user the import page once,
this data will not be overwritten.

## 2. tests

This bug flagged to me something that has been bothering me - we don't have coverage over what data is being sent to the API when a form is submitted. In the integration tests, we check that it goes to the right page after submission. On the unit tests, we check the individual functions on the typescript level. 
For regular forms I think we can rely on the form inputs doing their job and sending the body of the form to the backend. But for these 'custom form' pages like the OASyS import we aren't checking that the form sends the correct data from the nunjucks level all the way to the client. 

I've introduced a `verifyApplicationUpdate` that should do this, I'd be interested to know if others think it adds any value.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -156,7 +156,6 @@
       "riskToKnownAdult": "Medium",
       "riskToStaff": "Low",
       "lastUpdated": "2023-09-18",
-      "dateOfOasysImport": "2023-09-19",
       "additionalComments": "some comments"
     },
     "risk-to-others": {
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -114,19 +114,17 @@
     }
   },
   "risk-to-self": {
+    "oasys-import": { "oasysImportDate": "2023-09-21T15:47:51.430Z" },
     "vulnerability": {
       "vulnerabilityDetail": "example answer vulnerability",
-      "dateOfOasysImport": "2023-08-31",
       "confirmation": "confirmed"
     },
     "current-risk": {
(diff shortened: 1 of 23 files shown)


patrickjfl (approved):
Good spot, good fix and I definitely agree with the approach on testing custom form request bodies - big improvement :)

PR #239, opened 2023-10-03
Bugs with RoSH Summary page

Harriethw:
# Context

https://trello.com/c/dl2n6Ny1/343-rosh-rosh-risk-summary-from-oasys 

I noticed on the `dev` environment that we weren't seeing the risk ratings for data that looks like it had been retrieved (see before screenshot). I think this is because we were not using the right data model for what is returned from the API (`RoshRisks` instead of `RoshRisksEnvelope`). It is a bit hard to test locally because the wiremock endpoint doesn't seem to be working, but looking at the types on the API I think this is the right fix..

Because the values are now a nested object, I've had a go at pulling them out in the `response` function of the page. This might conflict with what you've been planning @patrickjfl 

I've also set the `back` button to return to the Task List.

### Before

Screenshot 2023-10-03 at 14 45 33

### After

Screenshot 2023-10-03 at 14 45 25

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -149,14 +149,17 @@
   "risk-of-serious-harm": {
     "oasys-import": { "oasysImportDate": "2023-09-21T15:47:51.430Z" },
     "summary": {
+      "dateOfOasysImport": "2023-09-21T15:47:51.430Z",
       "status": "retrieved",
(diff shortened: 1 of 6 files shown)


patrickjfl (approved):
great investigative work, and apologies for the mistake!

PR #242, opened 2023-10-09
Fix: RoSH risk summary not displaying on second viewing

Harriethw:
Screenshot 2023-10-09 at 14 22 47

# Context

Alex picked up in QA that the risk widget in the RoSH task displays correctly on first visit, but then becomes the 'unknown risk' version on second visits.

# Changes in this PR

The RoSH risk ratings are retrieved from the API and saved to our application data by the `oasysImport` page. This data is then retrieved from the application data used to populate the 'risk widget' by the Summary page.

The risk ratings were previously stored under the key for the Summary page, `summary`. When the user first lands on the Summary page we are able to see the widget, but when the user presses 'continue' the form page takes that `summary` key and uses it to overwrite the data that's there with it's own form data (see `save` function in the `applicationService`). Thus deleting the risk widget data.

This commit proposes a solution of saving the risk widget data under its own key on the application data. This means that it will not be overwritten by another form. The downside is that it effectively creates the idea of a `page` in the application data JSON that is not shown to the user. I think that this is mitigated by the pattern we are already using for 'add to list' pages, where we use one page to render data and another page key to store the data (e.g. `acct-data`). This widget is also a slight anomaly because a) it is not user-inputted data, but an exact copy from the API  b) The source of truth for application data should be the JSON itself, not the typescript form pages.

As we've had a few issues with this widget, I've also put an `e2e` test here https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-e2e/pull/30 
But this is a bit of a different test for our e2e's and would be interested in any feedback.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -148,7 +148,7 @@
   },
   "risk-of-serious-harm": {
     "oasys-import": { "oasysImportDate": "2023-09-21T15:47:51.430Z" },
-    "summary": {
+    "summary-data": {
       "dateOfOasysImport": "2023-09-21T15:47:51.430Z",
       "status": "retrieved",
       "value": {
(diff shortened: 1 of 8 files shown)


edavey (approved):
This looks like a good solution to me, but would like to know what @patrickjfl thinks...

PR #265, opened 2023-10-23
Feature/563 referrer can submit application

libuk:
# Context
Allow a referrer to submit an application.

# Changes in this PR
* Allow submission of an application
* Build application document on submission
* Mark check your answers task as incomplete if a previously answered task is changed
* Require confirmation checkbox to be checked before submission

Out of scope:
* Checking all answers are complete (incomplete applications can still be submitted)
* Confirmation page changes

Note: I wanted to include a new fixture for `applicationTranslatedDocument.json` to match the [other CAS services](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/cypress_shared/fixtures/applicationTranslatedDocument.json), but I ran out of time. Maybe this can be included in the task for creating the submitted application view.

### Before
Screenshot 2023-10-30 at 18 09 14

### After
Screenshot 2023-10-30 at 18 06 02

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/services/applicationService.ts
+++ b/server/services/applicationService.ts
@@ -51,6 +51,10 @@ export default class ApplicationService {
   async save(page: TaskListPage, request: Request) {
     const errors = page.errors()
 
+    const pageHasChangedSinceLastSave = (oldBody: Record<string, string> | undefined) => {
+      return !oldBody || !pageBodyShallowEquals(oldBody, page.body)
+    }
+
     if (Object.keys(errors).length) {
       throw new ValidationError<typeof page>(errors)
     } else {
@@ -73,7 +77,7 @@ export default class ApplicationService {
         application.data[checkYourAnswersTaskName] &&
         !(taskName === checkYourAnswersTaskName && pageName === checkYourAnswersPageName)
       ) {
-        if (!oldBody || !pageBodyShallowEquals(oldBody, page.body)) {
+        if (pageHasChangedSinceLastSave(oldBody)) {
           delete application.data[checkYourAnswersTaskName]
         }
       }
--- a/server/services/applicationService.ts
+++ b/server/services/applicationService.ts
@@ -1,10 +1,11 @@
 import type { Request } from 'express'
 import { AnyValue, Cas2Application as Application, Cas2Application } from '@approved-premises/api'
 import type { DataServices, GroupedApplications } from '@approved-premises/ui'
-import { getBody, getPageName, getTaskName } from '../form-pages/utils'
+import { getBody, getPageName, getTaskName, pageBodyShallowEquals } from '../form-pages/utils'
 import type { ApplicationClient, RestClientBuilder } from '../data'
 import { getApplicationSubmissionData, getApplicationUpdateData } from '../utils/applications/getApplicationData'
 import TaskListPage, { TaskListPageInterface } from '../form-pages/taskListPage'
+import CheckYourAnswers from '../form-pages/apply/check-your-answers/check-your-answers/checkYourAnswers'
 import { ValidationError } from '../utils/errors'
 
 export default class ApplicationService {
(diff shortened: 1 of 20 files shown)


server/utils/checkYourAnswersUtils.test.ts
```diff
@@ -17,7 +17,7 @@ const {
   getAnswer,
   summaryListItemForQuestion,
   checkYourAnswersSections,
-  getSectionsWithAnswers,
+  getSections,
```
edavey, line 20: interesting! I think we were trying for in `getSectionsWithAnswers` was a name which would communicate that we only want sections _which have answers_ rather than _sections and their answers_! i.e. sections without any answers (e.g. 'check your answers') are omitted from what's returned. Maybe `getSections()` is adequate rather than something more pedantic like `getSectionsContainingQuestions()`?

server/utils/checkYourAnswersUtils.test.ts
```diff
@@ -182,6 +182,44 @@ describe('checkYourAnswersUtils', () => {
...
+        },
+      ]
+
+      addPageAnswersToItemsArray({
+        items,
+        application: mockApplication,
+        task: 'confirm-eligibility',
+        pageKey: 'page1',
+        questions: mockedConfirmEligibilityQuestion,
+        checkYourAnswersPage: false,
```
edavey, line 222: do you think it might be clearer if we gave the "for the document" pathway equal status to the "for check your answers"?

i.e. rather than a boolean on the `checkYourAnswersPage` param, you do something like:

`outputFormat: 'checkYourAnswers' | 'document'`

server/utils/checkYourAnswersUtils.ts
```diff
@@ -42,28 +51,34 @@ export const getTaskAnswersAsSummaryListItems = (task: string, application: Appl
...
 
   const page = new PageClass({}, application)
 
   if (hasResponseMethod(page)) {
     const response = page.response()
     Object.keys(response).forEach(question => {
-      items.push(summaryListItemForQuestion(application, task, pageKey, { question, answer: response[question] }))
+      if (checkYourAnswersPage) {
+        items.push(summaryListItemForQuestion(application, task, pageKey, { question, answer: response[question] }))
+      } else {
```
edavey, line 79: see above, regarding the `checkYourAnswers` boolean argument - but I wonder if maintainers would find it clearer to see something like `else if 'document'` here?

server/services/applicationService.ts
```diff
@@ -59,10 +60,24 @@ export default class ApplicationService {
...
       application.data[taskName][pageName] = page.body
 
+      const checkYourAnswersTaskName = getTaskName(CheckYourAnswers)
+      const checkYourAnswersPageName = getPageName(CheckYourAnswers)
+
+      if (
+        application.data[checkYourAnswersTaskName] &&
+        !(taskName === checkYourAnswersTaskName && pageName === checkYourAnswersPageName)
+      ) {
+        if (!oldBody || !pageBodyShallowEquals(oldBody, page.body)) {
```
edavey, line 76: would it be helpful to wrap this rather opaque test in a local function with an intention-revealing name e.g. `anotherSectionHasSinceBeenEdited()` (if that's what this actually does?!)?

edavey (approved):
🏅 Excellent work, very easy to follow with helpful commentary.

I've made a couple of small suggestions to potentially add more clarity in a couple of places.

I think I've probably misunderstood what you're saying here about rendering the submitted `document` in the future, but just to reiterate that the code we'll use to display a `CAS2SubmittedApplication.document` needs to be completely independent of the UI's form pages/tasks/sections which are liable to change over time:

> We use buildDocument to create a snapshot of the application at
submission time. This will later be used to render the sections and
answers on the submitted application page.

server/utils/checkYourAnswersUtils.test.ts
libuk: I like this idea!

server/services/applicationService.ts
libuk: I definitely intended to pull this out into its own function, so I agree an a refactor would be useful here 👍🏾

libuk:
[Referrer can submit application](https://trello.com/c/vz1kkOsW/563-referrer-can-submit-application)

server/utils/checkYourAnswersUtils.test.ts
Harriethw: > i.e. rather than a boolean on the checkYourAnswersPage param, you do something like:

> outputFormat: 'checkYourAnswers' | 'document'

this sounds like a good idea, I've added a new commit introducing an `outputFormat`

server/services/applicationService.ts
Harriethw: I've done the renaming too... if I understand right it's checking whether a specific page has changed, so named accordingly

PR #274, opened 2023-11-01
(543) Assessor views a submitted application

edavey:
# Context

# Changes in this PR

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


integration_tests/tests/assess/view_submitted_application.cy.ts
```diff
@@ -0,0 +1,53 @@
...
+//    I want to view a particular submitted application
+//
+//  Scenario: follows an external link to the submitted application
+//    Given a submitted application exists
+//    And I am logged in as a NACRO assessor
+//    When I navigate to the url of the submitted application
+//    Then I see the assessor's read-only view of the submitted application
+//
+//  TODO
+//  Scenario: is redirected to the submitted application after logging in
```
Harriethw, line 13: would this scenario be our responsibility to test, as isn't this functionality handled by HMPPS Audit middleware? They just take whatever the redirect param is and take the user there...

edavey: good point!

edavey: I've removed that scenario

server/paths/assess.ts
```diff
@@ -0,0 +1,11 @@
+import { path } from 'static-path'
+
+const assessApplicationsPath = path('/assess/applications')
```
Harriethw, line 3: the test coverage is showing this file as having uncovered lines, whereas the other path files have full coverage... I can't work out how they are covered though 🤔

edavey: ah, thks I missed that -- I will take a look...

edavey:
[NACRO assessor views a submitted application](https://trello.com/c/e0eaNKqW/543-nacro-assessor-views-a-submitted-application)

server/paths/assess.ts
Harriethw: maybe it's just because it's never called in a unit test somewhere? looks like the others are, and when I put `expect(paths.submittedApplications.show({ id: submittedApplication.id })).toEqual(
        `/assess/applications/${submittedApplication.id}`,
      )` into the controller test it shows as covered... don't think this needs to hold things up. I wonder if we should put `/* istanbul ignore file */` at the top of these files actually, unless there is anything we think needs covering?

edavey: thank you 🙇 I've added that routing expectation into the controller test as it seems helpful...

patrickjfl (approved):
Looks great!

PR #287, opened 2023-11-07
Fix check for submitted application

Harriethw:
On applications that are not submitted the key will be present but the value will be `null`.


edavey (approved):
It would probably be a good idea to put a ticket in the backlog to backfill the tests to prevent a regression?

Harriethw:
@edavey oh good point. I've just pushed an update to the application factory, do you think that's enough? have we set `submittedAt` in any other tests/mocks for unsubmitted apps 🤔

Harriethw:
oh annoyingly Pact throws up an error if I set `submittedAt` on the `applicationFactory` to `null` 🤔 
Going to keep it as `undefined` for now and will create a ticket to look into it

PR #306, opened 2023-11-15
Silently handle errors for OASyS import

Harriethw:
If there is an error retrieving OASyS data, we do not want to surface that error to the user. We instead want to indicate that it was not possible to retrieve an oasys report at this time - the user is then given the option to proceed with the form and fill it in manually, or return another time.

This is to quickly fix an issue in development where if the API is not available, then the user is not able to proceed with form. In the future we may want to show a more helpful message based on the type of error received, e.g. 'Person has not been found', 'OASyS API is not available at this time'. We may want to consider what error to show if the user is unauthorized as well.

This is in line with how CAS1 handle errors
regarding OASyS [1], catching all instances of errors and allowing the user to proceed with a blank form.

[1] https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/utils/oasysImportUtils.ts#L42

### DIFF
--- a/server/form-pages/apply/risks-and-needs/risk-of-serious-harm/custom-forms/oasysImport.test.ts
+++ b/server/form-pages/apply/risks-and-needs/risk-of-serious-harm/custom-forms/oasysImport.test.ts
@@ -120,23 +120,14 @@ describe('OasysImport', () => {
     })
 
     describe('when oasys sections are not returned', () => {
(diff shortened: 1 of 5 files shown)


edavey (approved):
This seems like a good pragmatic quick fix. As you say, it would be good to develop this to distinguish between types of error e.g.

- there's no OASys for this user, please go and make one
- we couldn't connect to OASys, please try later or fill in the blanks manually

PR #310, opened 2023-11-17
Search for applicant by prison number

Harriethw:
# Context

We currently have the functionality to search (and add) a person by CRN. We received feedback that not all POMs would know how to get this ID, and the "prison number" (aka NOMIS ID, aka `nomsNumber` on the API) would be more appropriate. 
For the simplicity of the MVP we have decided to remove searching by CRN and only enable searching by prison number.

🚨 **This PR requires this API PR to merge first** https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1152 
**The end-to-end test is also ready to go:** https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-e2e/pull/39 

# Changes in this PR

I'm thinking that as there are several ways of referring to this ID, where the frontend displays it or knows about it we can call it `prison number` and where the API is responsible it will know it as `nomsNumber`. 

`findByCrn` in the person service becomes `findByPrisonNumber`  and we update the client and controller to handle it.

NOTE: 
* I need to confirm the exact design/copy on the template with UCD
* Currently when a person is found from the client we automatically go and create the application. There is an upcoming story to show some the information about the applicant before proceeding to create and save the application. 

### Before

Screenshot 2023-11-24 at 09 31 22

### After

Screenshot 2023-11-24 at 11 40 29

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


integration_tests/pages/apply/prisonNumberPage.ts
```diff
@@ -0,0 +1,12 @@
+import Page from '../page'
+
+export default class PrisonNumberPage extends Page {
```
patrickjfl, line 3: minor but maybe this would be better named `FindByPrisonNumberPage`?

server/utils/errors.ts
```diff
@@ -26,12 +26,3 @@ export class TaskListAPIError extends Error {
     this.field = field
   }
 }
-
-export class RestrictedPersonError extends Error {
```
patrickjfl, line 30: if the API is now returning a Forbidden error in these cases (LAO?), do we still want to handle that in a user friendly way in the UI?

patrickjfl (changes requested):
Looks good generally, nice work!

It looks like there are still a few references to CRNs in parts of the code that there shouldn't be, e.g. the confirm eligibility integration tests are looking for a 'crn' field in the prison number form.

Also maybe not for this PR but do we still want to surface CRNs in the applications dashboard, or move to prison number?

server/utils/errors.ts
Harriethw: We are showing the error `You do not have permission to access this prison number` rather than `CRN: is restricted` which is both more accurate and gives less information about the ID that's been searched. I don't know if it's over cautious but my thinking is that the user should only know that they are not allowed to access that ID, not necessarily the reason why, because that would allow someone with malicious intent to get more info than they need.

integration_tests/pages/apply/prisonNumberPage.ts
Harriethw: good spot, changed

Harriethw:
> It looks like there are still a few references to CRNs in parts of the code that there shouldn't be, e.g. the confirm eligibility integration tests are looking for a 'crn' field in the prison number form.

Thanks, thought I'd got rid of them, should be good now.

> Also maybe not for this PR but do we still want to surface CRNs in the applications dashboard, or move to prison number?

Good thinking. It looks like the prototype still has CRN on dashboard, I think if it needs changing it can be picked up in another ticket

server/utils/errors.ts
patrickjfl: right yeah, sounds sensible 👍

patrickjfl (approved):

PR #330, opened 2023-11-27
Confirm applicant details

Harriethw:
# Context

https://trello.com/c/QxtY2FOY/314-confirm-applicant 
Confirm applicant details, after searching for them using a prison number

needs this API PR to merge https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1189 

# Changes in this PR

This has involved moving the `create` application function back to the applications controller, rather than creating the application directly in the people controller. 

## Questions:

* I've put in some error handling on the applications controller `create` function:

```
      try {
        const application = await this.applicationService.createApplication(req.user.token, crn)

        res.redirect(paths.applications.show({ id: application.id }))
      } catch (error) {
        req.flash('errors', {
          prisonNumber: {
            text: 'There was an error creating the application, please try again',
            attributes: { 'data-cy-error-prisonNumber': true },
          },
        })
        req.flash('errorSummary', [
          { text: 'There was an error creating the application, please try again', href: '#prisonNumber' },
        ])
        res.redirect(paths.applications.new({}))
      }
```

So that any error from the server will be caught and the user taken back to enter a prison number again. 
There are a few different errors [we could get back from the API](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/c63a95fe13969be0ebd5ff0d69f070b470df958e/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt#L81):
* Offender doesn't exist
* Offender does not have noms Number
* User does not have permission to access
* Just a general error if a service or API went down?

The first three are unlikely to happen between the time that the user has searched for the applicant and clicked 'Confirm', but they could technically happen.
I'm proposing for now we don't need to surface those individual error messages to the user - we'll see them in Sentry if they do start to happen and can deal with them then?
I'm going to ask UCD team their thoughts on what the copy should be.

### Before

no screen

### After

Screenshot 2023-11-27 at 16 40 07

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/tests/apply/before_you_apply/confirm_eligibility/confirm_eligibility_task.cy.ts
+++ b/integration_tests/tests/apply/before_you_apply/confirm_eligibility/confirm_eligibility_task.cy.ts
@@ -37,7 +37,6 @@
 //    Then I should be able to 'Find by prison number'
 
 import Page from '../../../../pages/page'
-import FindByPrisonNumberPage from '../../../../pages/apply/findByPrisonNumberPage'
 import TaskListPage from '../../../../pages/apply/taskListPage'
 import ConfirmEligibilityPage from '../../../../pages/apply/confirmEligibilityPage'
 import IneligiblePage from '../../../../pages/apply/ineligiblePage'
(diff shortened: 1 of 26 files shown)


integration_tests/tests/apply/before_you_apply/confirm_eligibility/confirm_eligibility_task.cy.ts
```diff
@@ -64,18 +63,16 @@ context('Complete "Confirm eligibility" task in "Before you start" section', ()
...
     cy.signIn()
     //    And I have successfully found a person by prison number
-    const page = FindByPrisonNumberPage.visit(this.application.person.name)
-    page.getTextInputByIdAndEnterDetails('prisonNumber', person.nomsNumber)
+    cy.task('stubCreateApplication', { application: this.application })
+    cy.task('stubApplicationGet', { application: this.application })
+    cy.task('stubApplicationUpdate', { application: this.application })
 
-    page.clickSubmit()
+    // confirmationPage.clickSubmit()
```
patrickjfl, line 74: stray commented code here

patrickjfl (approved):
Big piece of work, looks great to me!

PR #360, opened 2023-12-11
Do not merge: Save Person on an Application

Harriethw:
# Context

API PR -> https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1226 

We are now requiring the full details of a person to be saved when creating a new application, rather than just the CRN.
This PR takes the fields required on the 'Confirm applicant details' screen and POSTs them via hidden form fields to the API.

We now also show these applicant details on Check Your Answers and the submitted application views for referrers and assessors

Screenshot 2023-12-13 at 11 59 09

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/@types/shared/models/Cas2NewApplication.ts
+++ b/server/@types/shared/models/Cas2NewApplication.ts
@@ -1,0 +1,17 @@
+/* generated using openapi-typescript-codegen -- do no edit */
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+export type Cas2NewApplication = {
+    crn: string;
+    nomsNumber: string;
+    name: string;
+    dateOfBirth: string;
+    nationality?: string;
+    sex?: string;
+    personStatus: 'InCustody' | 'InCommunity' | 'Unknown';
+    prisonName?: string;
+    pncNumber?: string;
+};
+
--- a/integration_tests/pages/apply/confirmApplicantPage.ts
+++ b/integration_tests/pages/apply/confirmApplicantPage.ts
@@ -13,7 +13,7 @@ export default class ConfirmApplicantPage extends Page {
     cy.get('.govuk-error-summary').contains('There was an error creating the application, please try again')
   }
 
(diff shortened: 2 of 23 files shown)


server/@types/shared/models/Cas2NewApplication.ts
```diff
@@ -0,0 +1,17 @@
+/* generated using openapi-typescript-codegen -- do no edit */
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+export type Cas2NewApplication = {
```
libuk, line 6: [question] Probably a comment for the API PR the name feels a bit misleading. Would `Cas2Applicant` be more appropriate? There doesn't seem to be any data here that's connected to the application itself.

libuk (approved):
Looks good 👍🏾

Just a comment on the new `Cas2NewApplication` entity.

server/@types/shared/models/Cas2NewApplication.ts
Harriethw: ah yeah I could see it being a big confusing, but it's a bit of a technicality on the API - although these fields refer to the Person, they are being saved on the Application database record. The `Cas2NewApplication` refers to like the packet of data needed to create a new Application, and all these fields get saved against the Application Entity (so they each refer to a column on the database for an Application). I feel like calling it `Cas2Applicant` might be confusing as well, because we'd be POSTing an `Applicant` to make an `Application`...

libuk: OK, I get what you mean. My thought was as we were previously sending the `crn` that adding these extra fields would be an extension of that and could be categorised as information about the applicant, but I understand how these values referring to columns in the application database may confuse things on the API side.

Happy to leave as is 👍🏾

Harriethw:
There is some ongoing conversation around how we save the applicant info, so putting this on hold https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1247

Harriethw:
closing as we've moved on and don't require this

PR #377, opened 2023-12-15
turn on error handling in environments that are not dev/local

Harriethw:
We want the error handling as close as possible to prod in our test environment, this ensures that error messages and stack traces are shared in environments that are NOT test or production.

https://trello.com/c/nsMHl851/667-tech-task-turn-on-error-handling-in-preprod-and-test


patrickjfl (approved):
LGTM!

PR #402, opened 2024-01-03
586 list submitted apps

Harriethw:
# Context

We need a page which lists all submitted applications so that a CAS2 administrator or assessor has a page they can go to debug things/find a lost application. It will also be used in e2e testing.

https://trello.com/c/RFzrEdvE/586-nacro-list-of-submitted-apps

**_We are aware that this could be a very expensive call when there are more than 100 submissions - we plan to look at filtering/pagination ASAP._**

# Changes in this PR

The table appears for assessors/admins at endpoint `assess/applications`
I've also added in a column for the prison number as this is the key ID which makes this table usable, and updated the existing table to referrers while I'm there.
## Screenshots of UI changes

new assessor view
Screenshot 2024-01-03 at 17 38 14

referrer view (with new prison number column)
Screenshot 2024-01-03 at 17 40 33

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


edavey (approved):
nice work! 

Should we make a couple of API tickets for i) pagination and ii) ordering?

I think this is also going to be expensive because of all the additional API calls needed to retrieve the Person record from Delius for each row and that we need to have another discussion about https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1247

PR #411, opened 2024-01-04
Allow access to preprod from anywhere

Harriethw:
It has been agreed that our production environment will be accessible to the public, protected by HMPPS Auth and 2FA. Our preproduction environment is currently only accesible via approved IPs, but should have parity with production.

This sets the IP `allowlist` to `null` which means that any IP will be allowed to access (as is the case currently with dev and test environments)


edavey (approved):
I believe that this is correct, thks!

PR #413, opened 2024-01-05
(754) Redirect all traffic to a new maintenance page

tahb:
# Context

https://trello.com/c/QqHdrhLZ

# Changes in this PR

We add a new environment variable which we can use to redirect users to in certain situations. The immediate one being we want to turn on production whilst waiting for our authority to operate to be approved. 

We may also have need of this in future in other ways if the service becomes unhealthy. In which case we can look to extend the copy of this new page to be more useful. Suggesting what people should do in the mean time, wait? or submit an application via another means.

I'm imagining that a second part to this work will be weaving through the logic to allow CAS2_ADMINS to bypass this in order to test the service.

- Testing this with an integration test was hard so I had to leave it with unit tests. More info in the commit.
- When adding a new static page there didn't seem to be an existing pattern so [following this conversation](https://mojdt.slack.com/archives/C05GEPLBRN3/p1704382966719859) I've proposed a new staticController which can be used for this and other pages like the cookie policy and privacy policy.

## Screenshots of UI changes

Screenshot 2024-01-05 at 10-16-13 Maintenance page

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


.env.example
```diff
@@ -3,3 +3,4 @@ API_CLIENT_SECRET=clientsecret
 COMMUNITY_ACCOMMODATION_API_SERVICE_NAME=approved-premises
 ENVIRONMENT=local
 APPROVED_PREMISES_ROOT_PATH=approved-premises
+MAINTENANCE_MODE=false
```
Harriethw, line 6: maybe slightly nitpicky of me, but would it be possible to have a more boolean-style name? `IN_MAINTENANCE_MODE` ?

server/app.ts
```diff
@@ -1,30 +1,30 @@
...
+import path from 'path'
 
 import createError from 'http-errors'
 import methodOverride from 'method-override'
 
-import nunjucksSetup from './utils/nunjucksSetup'
 import errorHandler from './errorHandler'
 import authorisationMiddleware from './middleware/authorisationMiddleware'
 import { metricsMiddleware } from './monitoring/metricsApp'
+import nunjucksSetup from './utils/nunjucksSetup'
```
Harriethw, line 13: Nice, I'd be up for introducing alphabetical sorting if the rest of the team didn't object, used to do it on previous project and found it easier to parse code

.env.example
tahb: Yeah sure thing, that's a bit clearer 👍

tahb:
@Harriethw I've renamed that environment variable and added: 
- a missing unit test file staticController
- a small piece to the readme to refer to the maintenance page as a utility we have available

Will merge this in at some point on Monday 👍

Harriethw:
@tahb nice one! 
Would this be on option in our disaster recovery plan too? https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4536664517/Disaster+recovery+-+CAS2

tahb:
Yes absolutely. I'll merge this and link it together.

PR #420, opened 2024-01-09
(754) Allow users with the CAS2_ADMIN role to bypass maintenance

tahb:
# Context

https://trello.com/c/QqHdrhLZ

# Changes in this PR

- The maintenance page can be bypassed by any user with role CAS2_ADMIN. This will allow the team to test the service is performing properly and debug any issues before disabling maintenance mode.
- The extraction of user roles was handled by the `dashboardController` so we went into a rabbit hole to try and bring that behaviour into the `populateCurrentUser` middleware to make it centralised.
- The data should probably be stored in res.session rather than res.session but that felt like one rabbit hole too far for this one

## Screenshots of UI changes

https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/assets/912473/2ca85435-3e87-4120-b7be-ea9805d59883

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/middleware/populateCurrentUser.test.ts
+++ b/server/middleware/populateCurrentUser.test.ts
@@ -1,0 +1,38 @@
+import { createMock } from '@golevelup/ts-jest'
+
+import type { Request, Response } from 'express'
+import type { UserService } from '../services'
+import type { UserDetails } from '../services/userService'
+import populateCurrentUser from './populateCurrentUser'
+
+describe('populateCurrentUser', () => {
+  const userService = createMock<UserService>()
+
+  const userDetailsObject: UserDetails = {
+    name: 'FakeUser',
+    displayName: 'Fake Display Name',
+  }
+
+  beforeEach(() => {
+    userService.getUser.mockReset()
+  })
+
+  it('set user details from the user service if not already set', async () => {
+    const request: jest.Mocked<Request> = createMock<Request>()
+    const response: jest.Mocked<Response> = createMock<Response>({
+      locals: {
+        user: {
+          name: 'FakeUser',
+          displayName: 'Fake Display Name',
+        },
+      },
+    })
+    const next = jest.fn()
+
+    userService.getUser.mockResolvedValue(userDetailsObject)
+
+    populateCurrentUser(userService)(request, response, next)
+
+    expect(response.locals.user).toEqual(userDetailsObject)
+  })
+})
--- a/server/middleware/populateCurrentUser.test.ts
+++ b/server/middleware/populateCurrentUser.test.ts
@@ -1,11 +1,33 @@
 import { createMock } from '@golevelup/ts-jest'
-
 import type { Request, Response } from 'express'
+import jwt from 'jsonwebtoken'
 import type { UserService } from '../services'
 import type { UserDetails } from '../services/userService'
 import populateCurrentUser from './populateCurrentUser'
 
 describe('populateCurrentUser', () => {
+  const createToken = (authorities: string[]) => {
+    const payload = {
+      user_name: 'USER1',
+      scope: ['read', 'write'],
+      auth_source: 'nomis',
+      authorities,
+      jti: 'a610a10-cca6-41db-985f-e87efb303aaf',
+      client_id: 'clientid',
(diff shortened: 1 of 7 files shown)


server/middleware/populateCurrentUser.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  const payload = {
+    user_name: 'USER1',
+    scope: ['read', 'write'],
+    auth_source: 'nomis',
+    authorities,
+    jti: 'a610a10-cca6-41db-985f-e87efb303aaf',
+    client_id: 'clientid',
+  }
+
+  return jwt.sign(payload, 'secret', { expiresIn: '1h' })
```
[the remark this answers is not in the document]
tahb: Test value so I've marked as a false positive.

server/services/userService.ts
```diff
@@ -1,7 +1,7 @@
-import { convertToTitleCase } from '../utils/utils'
 import type HmppsAuthClient from '../data/hmppsAuthClient'
+import { convertToTitleCase } from '../utils/utils'
 
-interface UserDetails {
+export interface UserDetails {
```
tahb, line 4: We'd like to use this in our new tests but I wonder if we should be using a factory instead. Though a factory for `users` felt right as it was a resource, I was less certain that a factory for an object such as `userDetails` worked. 

[Confusingly there is already an existing share type of the same name](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/02393e7de798a6b780e6c6480da0145bda0701a2/server/%40types/ui/index.d.ts#L87) which requires has all 4 fields (including `roles`). I don't think this works for us as we want to test that the setting of roles is done.

server/middleware/populateCurrentUser.test.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+describe('populateCurrentUser', () => {
+  function createResWithToken({ authorities }: { authorities: string[] }): Response {
```
Harriethw, line 22: v v minor but while we're here could change `createToken` and `createResWithToken` to arrow functions for style consistency, and have both within scope of `populateCurrentUser`

```suggestion

  const createToken = (authorities: string[]) => {
    const payload = {
      user_name: 'USER1',
      scope: ['read', 'write'],
      auth_source: 'nomis',
      authorities,
      jti: 'a610a10-cca6-41db-985f-e87efb303aaf',
      client_id: 'clientid',
    }

    return jwt.sign(payload, 'secret', { expiresIn: '1h' })
  }

  const createResWithToken = ({ authorities }: { authorities: string[] }): Response =>
    createMock<Response>({
      locals: {
        user: {
          token: createToken(authorities),
          roles: [],
        },
      },
    })
  
```

server/middleware/populateCurrentUser.ts
```diff
@@ -7,8 +8,9 @@ export default function populateCurrentUser(userService: UserService): RequestHa
     try {
       if (res.locals.user) {
         const user = res.locals.user && (await userService.getUser(res.locals.user.token))
```
Harriethw, line 10: I know this was in the template but this line confused me a bit. Can we remove the check for `res.locals.user &&` because it's already checked that in the enclosing if statement?

server/services/userService.ts
Harriethw: Not a typescript expert, but I think it makes sense to re-use this as a type rather than create a factory?
It seems like the `UserDetails` in `userService` is referring to a different thing than the shared `UserDetails` with the `roles` key? The `getUser` function wouldn't return an object with `roles`?

Harriethw (approved):
Looks great, thanks Tom! A good improvement to how we get roles

server/middleware/populateCurrentUser.test.ts
tahb: Yeah that's a fair spot. I copied the basis of these tests from dashboard controller and didn't even notice the styling change! The methods can be been in the same scope too, just got mixed up along the amount of test tweaking I went through 👍

PR #444, opened 2024-01-16
Allow full error trace in Nunjucks

patrickjfl:
When a global function in a template throws an error, we don’t see the full trace of the error, which makes diagnosing issues really hard.

This makes use of the [undocumented dev argument](mozilla/nunjucks#1430) which allows us to see the cause of the error in our logs and in Sentry.

[Copied from CAS1.](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/490/commits/8eed08b4e8698b0ad16bc285f731bc8b8f0cd3b2)

# Context

# Changes in this PR

### Before

Screenshot 2024-01-17 at 08 49 24

### After

Screenshot 2024-01-17 at 08 48 57

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/@types/nunjucks/index.d.ts
+++ b/server/@types/nunjucks/index.d.ts
@@ -1,0 +1,7 @@
+export default {}
+
+declare module 'nunjucks' {
+  export interface ConfigureOptions {
+    dev?: boolean | undefined
+  }
+}
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -57,6 +57,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     {
       autoescape: true,
       express: app,
+      dev: true, // This is set to true to allow us to see the full stacktrace from errors in global functions, otherwise it gets swallowed and tricky to see in logs
     },
   )
 


tahb (approved):
I see this is what CAS1 did but I'm feeling cautious. Configuring an app that's running in production to behave like dev might have other unexpected consequences. [The comment within nunjucks does suggests that all it does is change logging though](https://github.com/mozilla/nunjucks/blob/ea0d6d5396d39d9eed1b864febb36fbeca908f23/nunjucks/src/environment.js#L42).. [I tracked the origin way back and it does seem to have been added with the explicit purpose of improving error clarity](https://github.com/mozilla/nunjucks/commit/ea5557a8953cd0d189e233689fb4fb29139e0861).

I'm wary that this might somehow log too much info to the client or for any other part of Nunjucks to have plugged into this config (based on the generic name) which might cause some other dev only behaviour.

I ran out of time to try and observe better stack traces locally because ironically my stack seemed broken but I assume it works.

patrickjfl:
> I see this is what CAS1 did but I'm feeling cautious. Configuring an app that's running in production to behave like dev might have other unexpected consequences.
> 
> I'm wary that this might somehow log too much info to the client or for any other part of Nunjucks to have plugged into this config (based on the generic name) which might cause some other dev only behaviour.
> 
> [The comment within nunjucks does suggests that all it does is change logging though](https://github.com/mozilla/nunjucks/blob/ea0d6d5396d39d9eed1b864febb36fbeca908f23/nunjucks/src/environment.js#L42).. [I tracked the origin way back and it does seem to have been added with the explicit purpose of improving error clarity](https://github.com/mozilla/nunjucks/commit/ea5557a8953cd0d189e233689fb4fb29139e0861). And CAS1 have used it and no weird behaviour has been found yet. So on balance this is probably ok as everyone is a signed in user.
> 
> I ran out of time to try and observe better stack traces locally because ironically my stack seemed broken but I assume it works.

Good points. I guess on this basis we can try it and see how it goes.

It does work - I should have added some screenshots but will now.

PR #479, opened 2024-01-23
Add '/info' endpoint

Harriethw:
# Context

https://dsdmoj.atlassian.net/browse/CAS2-3 

This adds an `/info` endpoint for CAS2, which is required in order to share the prison codes for private beta roll out, as [required to have a tile on DPS](https://dsdmoj.atlassian.net/wiki/spaces/ALIGN/pages/4480959439/Adding+changing+services+on+the+common+header+footer+and+the+homepage#Establishment-specific-rollouts).

It should return 

```
{
  "activeAgencies": [ ],
}
```

# Changes in this PR

The `activeAgencies` list will in the near future be set as a list of the prison codes we're rolling out to. An empty list is the preferred state for a service that is not yet rolled out.

When we are ready to launch private beta, we can update this list with the codes required, and DPS will be able to pick the information up within an hour.

When looking at the [hmpps-typescript-template repo](https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/middleware/setUpHealthChecks.ts#L24), this `/info` endpoint also has a productId, as well as info about git and version. As this is a public repo, I'm not going to add any information except the `activeAgencies` required, as there could be risks around exposing it.

The template has the route setup in `setUpHealthCheck` but I thought it made more sense to have this in a separate file, as this is not related to the `/health` endpoint.

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/integration_tests/tests/productInfo.cy.ts
+++ b/integration_tests/tests/productInfo.cy.ts
@@ -1,0 +1,13 @@
+context('Product info', () => {
+  context('Info provided', () => {
+    beforeEach(() => {
+      cy.task('reset')
+      cy.task('stubAuthPing')
+      cy.task('stubTokenVerificationPing')
+    })
+
+    it('Returns an empty list of active agencies', () => {
+      cy.request('/info').its('body.activeAgencies').should('be.empty')
+    })
+  })
+})
--- a/server/app.ts
+++ b/server/app.ts
@@ -26,6 +26,7 @@ import setUpWebRequestParsing from './middleware/setupRequestParsing'
 import { Controllers } from './controllers'
 import routes from './routes'
 import type { Services } from './services'
+import setUpProductInfo from './middleware/setUpProductInfo'
 
 export default function createApp(controllers: Controllers, services: Services): express.Application {
   const app = express()
@@ -40,6 +41,7 @@ export default function createApp(controllers: Controllers, services: Services):
 
   app.use(metricsMiddleware)
   app.use(setUpHealthChecks())
+  app.use(setUpProductInfo())
   app.use(setUpWebSecurity())
   app.use(setUpWebSession())
   app.use(setUpWebRequestParsing())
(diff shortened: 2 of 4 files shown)


integration_tests/tests/productInfo.cy.ts
```diff
@@ -0,0 +1,13 @@
+context('Product info', () => {
+  context('Info provided', () => {
+    beforeEach(() => {
+      cy.task('reset')
+      cy.task('stubAuthPing')
+      cy.task('stubTokenVerificationPing')
+    })
+
+    it('Product ID is visible', () => {
```
tahb, line 9: I guess this from the test in the template we copied from?

tahb (approved):
LGTM

I suspect the product Id is related to an initiative from MoJ to compile a service catalogue. [I ran into this over on the dps-project-bootstrap repo when trying to make an unrelated update](https://github.com/ministryofjustice/dps-project-bootstrap/pull/562/commits/6bdc93207bf37ae8919561742536fd691b470fbf). I don't think we add it now but if this is the same thing this link might be useful for a future person to read.

Harriethw:
>I suspect the product Id is related to an initiative from MoJ to compile a service catalogue. [I ran into this over on the dps-project-bootstrap repo when trying to make an unrelated update](https://github.com/ministryofjustice/dps-project-bootstrap/pull/562/commits/6bdc93207bf37ae8919561742536fd691b470fbf).

@tahb Thanks for adding the link - I originally put the product ID in, but then as [activities-management](https://github.com/ministryofjustice/hmpps-activities-management/blob/1f66bac25680e40597853614bd721a735740e558/server/middleware/setUpHealthChecks.ts#L20) didn't include it, and I wasn't sure if it should be 'secret', I've left it out. 
I think this is the service to get the ID's https://developer-portal.hmpps.service.justice.gov.uk/products (only available via MoJ macbook).

I'll add this context to the commit message.

integration_tests/tests/productInfo.cy.ts
Harriethw: corrected!

PR #493, opened 2024-01-29
Adds cancel buttons to ACCT and current offence pages

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/CAS2-173

# Changes in this PR

Adds cancel buttons for ACCT and current offence pages. The current offence button is a bit delicate because that is the first page the user lands on from the task list when they first begin the task. In that case, I assume we don't want to render a cancel button - we only want to after an offence has been added, when there is a 'Current offences' (list) page to redirect them to. So I have added an attribute on that page to check if there is data for the task on the application.

### Before

Screenshot 2024-01-29 at 14 33 36

Screenshot 2024-01-29 at 14 34 48

### After

Screenshot 2024-01-29 at 14 34 01

Screenshot 2024-01-29 at 14 35 18

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/form-pages/apply/offence-and-licence-information/current-offences/custom-forms/currentOffenceData.ts
+++ b/server/form-pages/apply/offence-and-licence-information/current-offences/custom-forms/currentOffenceData.ts
@@ -50,7 +50,7 @@ export default class CurrentOffenceData implements TaskListPage {
 
   offenceCategories: Array<SelectItem>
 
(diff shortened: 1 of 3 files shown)


libuk (approved):
This approach makes sense to me!

PR #482, opened 2024-01-24
Chore/run e2e tests from within the repo

tahb:
# Context

https://dsdmoj.atlassian.net/browse/CAS2-140?atlOrigin=eyJpIjoiOGYwMzUwYmJkMWZkNDY3NGEyNGUyZTM3ZGFkYmIzYjciLCJwIjoiaiJ9

It’s important to note that e2e tests are required to pass for deploying the frontend to preprod and prod.

We found that it can be difficult to coordinate the frontend changes (a pull request in 1 repo) alongside the changes to the e2e tests (a pull request in another repo).

Both must be worked on and tested locally from different repos. We also have double the admin of proposing and reverting two pull requests, rather than just one.

We’d like to move these tests from a separate repo to the same repo.

# Changes

- We install playwright and copy over the e2e tests from the separate repo.
- We add a very simple health check test to help quickly verify the e2e tests are set up correctly. We could drop this if needed?

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Are any changes required to the e2e tests?
- [ ] If you've added a new route, have you added a new
  `auditEvent`? (see `server/routes/apply.ts` for examples)
- [ ] Are there environment variables or other infrastructure configuration which needs to be included in this release?
- [ ] Are there any data migrations required. Automatic or manual?
- [ ] Does this rely on changes being deployed to the CAS API?

## Post merge checklist

Once we've merged it will be auto-deployed to the dev environment.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-community-accommodation-tier-2-ui).

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #495, opened 2024-01-29
Run new e2e tests in the pipeline

tahb:
# Context

# Changes in this PR

We want to run e2e tests from within the repo, rather than the dedicated repo. [Now we've moved the tests over](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/482) we can introduce it into the pipeline.

So that we don't block the pipeline getting this set up, the new job is optional but allows us to see the result so we know if it's working.

We keep the old tests around a little longer which must still pass for us to release. 

Next we'll switch this over and remove the old job.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


.circleci/config.yml
```diff
@@ -193,6 +193,34 @@ jobs:
...
+          command: |
+            git clone https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui.git .
+      - run:
+          name: Update npm
+          command: 'npm install -g npm@latest'
+      - node/install-packages
+      - run:
+          name: E2E Check
+          command: |
+            npm run test:e2e
```
tahb, line 212: I'm hoping I can just switch it to this instead of `npm run test` 🤞

.circleci/config.yml
```diff
@@ -193,6 +193,34 @@ jobs:
...
 
+  e2e_environment_test:
+    docker:
+      - image: mcr.microsoft.com/playwright:v1.40.0-focal
+    circleci_ip_ranges: true # opt-in to jobs running on a restricted set of IPs
+    steps:
+      - run:
+          name: Clone E2E tests
+          command: |
+            git clone https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui.git .
```
tahb, line 204: I considered whether cloning it was still the right approach since this job is already operating from the context of this repo now, and for it to behave more like `unit_test`

I think sticking with the same approach of the old repo is likely the simplest approach to start with as it's similar to what we know works. If it does work, then we can decide if we want to change how it runs to be more like `unit_test`.

.circleci/config.yml
libuk: Would we need to [provide env variables](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/.circleci/config.yml#L67) like CAS3?

.circleci/config.yml
libuk: We don't seem to be doing any [cloning in CAS3](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/.circleci/config.yml), so I'm not sure what the result of this would be, but no harm in testing.

libuk (approved):
Happy for this to go in and start testing, but had a couple of questions.

.circleci/config.yml
tahb: This has made me think that I could try to add another job in for the non-cloning path and see if it works!

.circleci/config.yml
tahb: I'm hoping that [the same environment variables we set on the CircleCI project for the old job](https://app.circleci.com/settings/project/github/ministryofjustice/hmpps-community-accommodation-tier-2-ui/environment-variables) will be there in the same way 🤞

.circleci/config.yml
tahb: [It worked without cloning quite easily](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-community-accommodation-tier-2-ui/3208/workflows/d3548fad-7978-46ce-8b7e-584c3106c192/jobs/12782). I worked out how to test it before pushing too. I'm going to reforge this PR to basically swap out the insides of the same job.

tahb:
@libuk We've gone with the non-cloning option now, thanks for the push! I've left the WIP in that runs the e2e job as part of this pipeline so you can see it passing, i'll remove that before merging.

PR #521, opened 2024-02-09
fix assessor e2e tests to be person-agnostic

Harriethw:
As the dev environment has seeded data with test
names, the test has to be neutral on the personal
details of the application it picks, and just
check for known elements.

### DIFF
--- a/e2e-tests/tests/03_assess.spec.ts
+++ b/e2e-tests/tests/03_assess.spec.ts
@@ -2,12 +2,12 @@ import { Page, expect } from '@playwright/test'
 import { signInAsAssessor, updateStatus, viewSubmittedApplication } from '../steps/assess'
 import { test } from '../test'
 
-test('view a submitted application as an admin', async ({ page, assessorUser, person }) => {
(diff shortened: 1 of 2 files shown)


e2e-tests/tests/03_assess.spec.ts
```diff
@@ -2,12 +2,12 @@ import { Page, expect } from '@playwright/test'
...
-test('view a submitted application as an admin', async ({ page, assessorUser, person }) => {
+test('view a submitted application as an admin', async ({ page, assessorUser }) => {
   await signOut(page)
   await signInAsAssessor(page, assessorUser)
-  await viewSubmittedApplication(page, person)
-  await updateStatus(page, person)
-  await expect(page.locator('h1')).toContainText(person.name)
+  await viewSubmittedApplication(page)
+  await updateStatus(page)
+  await expect(page.locator('.moj-timeline__title').first()).toContainText('More information requested')
```
Harriethw, line 10: we have to have one `expect` in the test function here or else the linter complains, otherwise this line should live in the `updateStatus` function.

patrickjfl (approved):

PR #536, opened 2024-02-21
(210) Add e2e tests for reports

libuk:
[Jira](https://dsdmoj.atlassian.net/browse/CAS2-210?atlOrigin=eyJpIjoiNDU0MDI0YzVkZWU0NGFkYWExNmJhN2E1YjYxOTllY2UiLCJwIjoiaiJ9)

# Context
Test that a user can download reports and that the spreadsheet contains the expected columns.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


e2e-tests/steps/reports.ts
```diff
@@ -0,0 +1,72 @@
...
+export const confirmColumnNames = async (reportType: ReportType, path: string) => {
+  const workbook = new Excel.Workbook()
+
+  await workbook.xlsx.readFile(path).then(() => {
+    const sh = workbook.getWorksheet('Sheet0')
+
+    const headerCells: CellValue[] = []
+    sh.getRow(1).eachCell(cell => headerCells.push(cell.value))
+
+    expect(headerCells.length === reportMetaData[reportType].columnNames.length).toBe(true)
```
libuk, line 66: This assertion means that if the report is updated on the API side without an update on the FE side this test will fail. I was in two minds about this scenario blocking the FE pipeline, but let me know your thoughts.

e2e-tests/steps/reports.ts
```diff
@@ -0,0 +1,72 @@
...
+export const confirmColumnNames = async (reportType: ReportType, path: string) => {
+  const workbook = new Excel.Workbook()
+
+  await workbook.xlsx.readFile(path).then(() => {
+    const sh = workbook.getWorksheet('Sheet0')
+
+    const headerCells: CellValue[] = []
+    sh.getRow(1).eachCell(cell => headerCells.push(cell.value))
+
+    expect(headerCells.length === reportTypeMetaData[reportType].columnNames.length).toBe(true)
```
libuk, line 66: This assertion means that if the report is updated with a new column on the API without an update on the FE this test will fail. I was in two minds about this scenario blocking the FE pipeline, but let me know your thoughts.

patrickjfl: Is this in place to make sure we're not leaking data unexpectedly, or to make sure we keep our E2E tests up to date? If it's the latter then I'm not sure the best way to control that is with tight tests, at the expense of blocking the FE pipeline as you say. Not a major deal though.

e2e-tests/steps/reports.ts
```diff
@@ -0,0 +1,67 @@
...
+    ],
+    callToAction: "Download 'Submitted applications' report",
+  },
+  applicationStatusUpdates: {
+    columnNames: ['eventId', 'applicationId', 'personCrn', 'personNoms', 'updatedBy', 'updatedAt', 'newStatus'],
+    callToAction: "Download 'Application status updates' report",
+  },
+  unsubmittedApplications: {
+    columnNames: ['applicationId', 'personCrn', 'personNoms', 'startedBy', 'startedAt'],
+    callToAction: "Download 'Un-submitted applications' report",
```
patrickjfl, line 27: I guess this is coming from the API, but 'un-submitted' applications feels weird content wise.

patrickjfl (approved):

patrickjfl:
Approved but it would be good to test these locally - where can I find the local MI user credentials?

libuk:
@patrickjfl 

> Is this in place to make sure we're not leaking data unexpectedly, or to make sure we keep our E2E tests up to date? If it's the latter then I'm not sure the best way to control that is with tight tests, at the expense of blocking the FE pipeline as you say. Not a major deal though.

It was for the latter, I'll remove as it's probably overkill 👍🏾 

> Approved but it would be good to test these locally - where can I find the local MI user credentials?

You can find the credentials in 1password

e2e-tests/steps/reports.ts
libuk: This matches the content in our FE, so we can change this at any point. I'll run it past Chanel 👍🏾

patrickjfl:
> @patrickjfl
> 
> > Is this in place to make sure we're not leaking data unexpectedly, or to make sure we keep our E2E tests up to date? If it's the latter then I'm not sure the best way to control that is with tight tests, at the expense of blocking the FE pipeline as you say. Not a major deal though.
> 
> It was for the latter, I'll remove as it's probably overkill 👍🏾
> 
> > Approved but it would be good to test these locally - where can I find the local MI user credentials?
> 
> You can find the credentials in 1password

Ah thanks, wasn't sure if the dev ones would work locally. The tests are passing locally for me 👍

e2e-tests/steps/reports.ts
libuk: Leaving as is for now, but the conversation is here: https://mojdt.slack.com/archives/C04T6SC31HA/p1709137791628899

PR #558, opened 2024-03-06
Document our approach to removing old answers on the application

tahb:
This came up today as we discussed implementing another type of this change. We don't have a good way to automate this so it requires a human to remember which risks it being forgotten.

- We could write this in a checklist but it probably doesn't happen enough so we'd likely not notice it on the rare occasion it would be relevant. A similar thing could happen with Jira card checklist
- We could try and automate this somehow by redesigning how the question framework so it requires us to be more explicit about the relationship around dependencies. Such that the adding of the question and it's sub-questions is expressed explicitly so it could be automatically torn down. We're not sure how this would work, it would likely require more time than the couple of months we have left

# Changes in this PR

- document the problem within the repo so we have a easy thing to link to and improve the probability that a new developer reads through when joining

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/doc/how-to/removing_old_answers_on_an_application.md
+++ b/doc/how-to/removing_old_answers_on_an_application.md
@@ -1,0 +1,42 @@
+
+# Removing old answers on an Application
+
+- The application form is built using the inherited [task list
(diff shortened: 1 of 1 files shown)


patrickjfl (approved):
thanks for documenting this Tom 👍

PR #582, opened 2024-03-18
(CAS2-290) Validate the referer header used in backlinks and redirects

Gweaton:
# Context

[Jira ticket 290](https://dsdmoj.atlassian.net/browse/CAS2-290)

# Changes in this PR

We're currently not validating the `referer` header, which theoretically allows for "Reflected XSS attacks", as a `referer` could be passed in the header of a GET request. Although this is currently blocked by our CSP settings, we've been recommended to fix this by the pen test report.

This introduces a `validateReferer` function that checks the `referer` comes from our own URL, read from the ingress in the config. We've opted to revert to the root URL (`/`) for the back link otherwise.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/views/applications/taskList.njk
+++ b/server/views/applications/taskList.njk
@@ -9,10 +9,10 @@
 {% endblock %}
 
 {% block beforeContent%}
-  {% if referrer %}
+  {% if referer %}
     {{ govukBackLink({
         text: "Back",
-        href: referrer
+        href: referer
     }) }}
   {% endif %}
 {% endblock %}
--- a/feature.env
+++ b/feature.env
@@ -9,3 +9,4 @@ SYSTEM_CLIENT_ID=clientid
 SYSTEM_CLIENT_SECRET=clientsecret
 APPROVED_PREMISES_API_URL=http://localhost:9999
 REDIS_PORT=6380
+INGRESS_URL=http://localhost:3007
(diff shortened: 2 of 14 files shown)


libuk (approved):
Nice work, thanks for the clear commit messages!

server/views/applications/taskList.njk
```diff
@@ -9,10 +9,10 @@
 {% endblock %}
 
 {% block beforeContent%}
-  {% if referrer %}
+  {% if referer %}
     {{ govukBackLink({
         text: "Back",
-        href: referrer
+        href: referer | validateReferer
```
Harriethw, line 15: do you think we should leave the validation of the `referer` to the controllers? In this case this `referer` has already been validated by `showMissingRequiredTasksOrTaskList`. I feel like if we are expecting to also validate it in the templates it could open door to double-validation like this. Also feels like the nunjucks should be able to trust anything they've been passed from the controller?

Gweaton: ah yes, good point - we don't need it here. I think I was a bit confused about where that `referer` was coming from early on and didn't update it when I figured it out. Can remove this one now.

Gweaton: And yeah, I think that's a good rule of thumb

PR #620, opened 2024-04-11
Update getApplications to use isSubmitted query param

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=CAS2-375

# Changes in this PR

Updates the fetch applications methods for referrers to use the isSubmitted query param, to separate calls for in progress and submitted applications.

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -26,7 +26,15 @@ export default class ApplicationsController {
 
   index(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const applications = await this.applicationService.getAllForLoggedInUser(req.user.token)
+      const inProgressApplications = await this.applicationService.getApplicationsBySubmissionStatusForLoggedInUser({
+        token: req.user.token,
+        isSubmitted: false,
+      })
+
+      const submittedApplications = await this.applicationService.getApplicationsBySubmissionStatusForLoggedInUser({
+        token: req.user.token,
+        isSubmitted: true,
+      })
 
       const { errors, errorSummary, userInput } = fetchErrorsAndUserInput(req)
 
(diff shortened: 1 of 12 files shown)


server/controllers/apply/applicationsController.ts
```diff
@@ -26,15 +26,24 @@ export default class ApplicationsController {
 
   index(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const applications = await this.applicationService.getAllForLoggedInUser(req.user.token)
+      const inProgressApplications = await this.applicationService.getApplicationsBySubmissionStatusForLoggedInUser({
```
Harriethw, line 29: I think this new param is intended to reduce volume of applications and calls made... 
would it be possible to just make one call to the API, and pass in the `isSubmitted` param from the UI path?
e.g. when I'm on the in progress tab we just make one call to the API with `isSubmitted` as false, and it just returns those applications. And vice versa for submitted tab

Harriethw: I think CAS1 do something similar with their dashboard function the applications controller

patrickjfl: hmmmm, actually I don't think that is possible, because the hash fragment isn't available in the controller...

patrickjfl:
Converting to draft as it's not possible to make distinct calls to feed in data for tabs within the govukTabs component. However, we will probably want to use the client/service changes for the prisons dashboard work.

PR #629, opened 2024-04-15
hot fix: set prison dashboard feature flag to be based on environment

Harriethw:
We discovered that .env/helm env variables are types of strings rather than booleans - that meant that if the .env variable was set to `false`, in javascript land that was becoming a string `'false'` and always evaluating as true.

This removes the flag as an environment variable and just sets the flag based on what the environment is - this should still allow us to easily turn it off/on and remove it once we're done.


patrickjfl (approved):
LGTM!

PR #619, opened 2024-04-10
CAS2-392: add e2e test for viewing application within user's prison

Harriethw:
This [depends on work in the API to introduce](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1735) permissions for users to view any submitted applications within their own prison. 

We are using a hardcoded seeded UUID for the application, but we will shortly be expanding the test to start from viewing a 'prison dashboard', from which the user will click through to the application.

I think currently this step still sits within a referrer's 'apply' journey so I haven't created a new spec for it, but I think 
as the dashboard requirements grow it may make sense to split out. 

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Are any changes required to the e2e tests?
- [ ] If you've added a new route, have you added a new
  `auditEvent`? (see `server/routes/apply.ts` for examples)
- [ ] Are there environment variables or other infrastructure configuration which needs to be included in this release?
- [ ] Are there any data migrations required. Automatic or manual?
- [ ] Does this rely on changes being deployed to the CAS API?

## Post merge checklist

Once we've merged it will be auto-deployed to the dev environment.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-community-accommodation-tier-2-ui).

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #652, opened 2024-04-23
Add applications table to prison dashboard

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=CAS2-391

# Changes in this PR

### Before

Screenshot 2024-04-23 at 16 18 10

### After

Screenshot 2024-04-23 at 16 17 34

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/e2e-tests/steps/apply.ts
+++ b/e2e-tests/steps/apply.ts
@@ -91,7 +91,7 @@ export const submitApplication = async (page: Page) => {
 
 export const viewSubmittedApplication = async (page: Page, name: string) => {
   await page.goto('/applications#submitted')
-  await expect(page.locator('h1')).toContainText('Short-Term Accommodation (CAS-2) applications')
+  await expect(page.locator('h1')).toContainText('Your CAS-2 applications')
   await page.getByRole('link', { name }).first().click()
   await expect(page.locator('h1')).toContainText(name)
   await expect(page.locator('h2').first()).toContainText('Application history')
--- a/e2e-tests/steps/apply.ts
+++ b/e2e-tests/steps/apply.ts
@@ -103,4 +103,12 @@ export const addNote = async (page: Page) => {
   await page.getByTestId('submit-button').click()
   await expect(page.locator('h2').first()).toContainText('Success')
   await expect(page.locator('.moj-timeline__description').first()).toContainText(note)
+  await expect(page.locator('.moj-timeline__title').first()).toContainText('Note')
+}
+
+export const viewApplicationFromPrisonDashboard = async (applicationId: string, page: Page, name: string) => {
+  await page.goto(`/applications/prison`)
+  await page.locator(`a[href="/applications/${applicationId}/overview"]`).click()
+  await expect(page.locator('h1')).toContainText(name)
+  await expect(page.locator('h2').first()).toContainText('Application history')
 }
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -118,7 +118,7 @@ const token = (authorities: Array<string>) =>
     },
   })
 
-const stubUser = (name: string) =>
+const stubUser = (name: string, activeCaseLoadId?: string) =>
   stubFor({
     request: {
       method: 'GET',
@@ -134,6 +134,7 @@ const stubUser = (name: string) =>
         username: 'USER1',
         active: true,
         name,
+        activeCaseLoadId,
       },
     },
   })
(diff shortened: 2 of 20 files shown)


server/data/applicationClient.ts
```diff
@@ -33,6 +33,11 @@ export default class ApplicationClient {
     return (await this.restClient.get({ path: paths.applications.index.pattern })) as Array<Cas2ApplicationSummary>
   }
 
+  async getAllByPrison(prisonCode: string): Promise<Array<Cas2ApplicationSummary>> {
+    const path = `${paths.applications.index({})}?prisonCode=${prisonCode}`
```
Harriethw, line 37: We also need to pass in the `isSubmitted=true` here, as we are only showing submitted applications currently (If this isn't passed in it will return all applications for the prison, which would lead to user clicking through to an in progress one and getting forbidden, because we haven't set up that user journey/permission yet)

server/utils/applicationUtils.ts
```diff
@@ -31,6 +31,18 @@ export const submittedApplicationTableRows = (
   })
 }
 
+export const prisonDashboardTableRows = (applications: Array<Cas2ApplicationSummary>): Array<TableRow> => {
+  return applications.map(application => {
+    return [
+      nameAnchorElement(application.personName, application.id),
+      textValue(application.nomsNumber),
+      textValue(application.createdByUserName),
+      textValue(DateFormats.isoDateToUIDate(application?.hdcEligibilityDate, { format: 'medium' })),
```
Harriethw, line 40: is `application` here always defined, so doesn't need the `?` 
`hdcEligibilityDate` looks like it is nullable though, do we need an example in the unit test of one without a date?

integration_tests/mockApis/auth.ts
```diff
@@ -134,6 +134,7 @@ const stubUser = (name: string) =>
         username: 'USER1',
         active: true,
         name,
+        activeCaseLoadId,
```
Harriethw, line 137: ah nice 👍

e2e-tests/steps/apply.ts
```diff
@@ -91,7 +91,7 @@ export const submitApplication = async (page: Page) => {
 
 export const viewSubmittedApplication = async (page: Page, name: string) => {
   await page.goto('/applications#submitted')
-  await expect(page.locator('h1')).toContainText('Short-Term Accommodation (CAS-2) applications')
+  await expect(page.locator('h1')).toContainText('Your CAS-2 applications')
```
Harriethw, line 94: whoops, looks like I never merged this e2e test for viewing prison application https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/619 
do you want to just do the whole journey in this PR, starting from the dashboard instead of going directly to the seeded application?

Harriethw: although I guess to ensure it's an application made by another user we will still need to use the seeded application...I wonder if there's a way to find a url with that ID on a page, in order to click it...

server/data/applicationClient.ts
patrickjfl: aha, thanks!

server/utils/applicationUtils.ts
patrickjfl: (and if so, how should we handle that?)

Harriethw (changes requested):
looks really good, just some small changes 👍

e2e-tests/steps/apply.ts
patrickjfl: hmmmm yeah, I remember this being tricky, but I'll have a look...

server/utils/applicationUtils.ts
Harriethw: actually there may be some submitted applications without the date that were submitted before we introduced that field

e2e-tests/steps/apply.ts
Harriethw: if it's not possible we could have two e2e tests:
1. to check they can see a dashboard
2. to check they can visit  {seeded_application_id}/overview and leave a note

Harriethw (approved):
Looks good, thank you!

e2e-tests/steps/apply.ts
```diff
@@ -103,4 +103,12 @@ export const addNote = async (page: Page) => {
   await page.getByTestId('submit-button').click()
   await expect(page.locator('h2').first()).toContainText('Success')
   await expect(page.locator('.moj-timeline__description').first()).toContainText(note)
+  await expect(page.locator('.moj-timeline__title').first()).toContainText('Note')
+}
+
+export const viewSubmittedApplicationOverview = async (applicationId: string, page: Page, name: string) => {
```
Harriethw, line 109: ooh maybe small thing but could this be renamed to be more specific? `viewApplicationFromPrisonDashboard` or something?

PR #657, opened 2024-04-24
CAS2-416: Allow Licence CA's to see prison dashboard

Harriethw:
# Context

CAS2 currently is setup for users with the POM
role to be able to create/read applications. We
have since found that the ROLE_POM is given out to
a broad range of users who may not be POMs - The
[LICENCE_CA role](https://dsdmoj.atlassian.net/wiki/spaces/CVL/pages/3540320445/Users+and+Roles) is more in line with who our
users are, and is also used by HDC services to
perform similar actions that we require. We are
therefore going to add in the LICENCE_CA role and
remove the POM role permissions in stages.

This is the first stage - to allow them to view submitted applications for their prisons.

# Changes in this PR

Having done the work on [the API to allow read
access to LICENCE_CA]( https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1786) this PR allows the users to see the prison dashboard tile.

A User with the LICENCE_CA role has been created for AP Tools locally, and I've setup one for the dev/test environment under our 1Password vault as 'DEV: CAS2 Licence CA user'

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/utils/userUtils.test.ts
+++ b/server/utils/userUtils.test.ts
@@ -11,6 +11,11 @@ describe('userUtils', () => {
       expect(sectionsForUser(['ROLE_POM'])).toEqual(expected)
     })
 
+    it('should return the prison dashboard for a Licence CA', () => {
+      const expected = [sections.prisonDashboard]
+      expect(sectionsForUser(['ROLE_LICENCE_CA'])).toEqual(expected)
+    })
+
     it('should return correct sections for an admin', () => {
       const expected = [sections.submittedApplications]
       expect(sectionsForUser(['ROLE_CAS2_ADMIN'])).toEqual(expected)
--- a/server/controllers/apply/applicationsController.test.ts
+++ b/server/controllers/apply/applicationsController.test.ts
@@ -676,18 +676,39 @@ describe('applicationsController', () => {
   })
 
   describe('prisonDashboard', () => {
-    it('renders the prison dashboard page', async () => {
-      response.locals.user = { activeCaseLoadId: '123' }
-      const prisonApplications = applicationSummaryFactory.buildList(5)
+    describe('when the user has the POM role', () => {
+      it('renders the prison dashboard page with isUserPom as true', async () => {
+        response.locals.user = { activeCaseLoadId: '123', roles: ['ROLE_POM'] }
+        const prisonApplications = applicationSummaryFactory.buildList(5)
 
-      applicationService.getAllByPrison.mockResolvedValue(prisonApplications)
+        applicationService.getAllByPrison.mockResolvedValue(prisonApplications)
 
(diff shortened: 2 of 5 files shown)


server/utils/userUtils.test.ts
```diff
@@ -21,6 +26,11 @@ describe('userUtils', () => {
         const expected = [sections.referral, sections.newReferral]
         expect(sectionsForUser(['ROLE_POM'])).toEqual(expected)
       })
+      it('should not return the prison dashboard for a Licence CA', () => {
+        config.flags.isPrisonDashboardEnabled = false
+        const expected: string[] = []
+        expect(sectionsForUser(['ROLE_LICENCE_CA'])).toEqual(expected)
+      })
```
patrickjfl, line 33: would it make sense to remove the feature flag before merging this branch?

patrickjfl (approved):
Looks OK to me. Are we going to end up removing the POM role altogether though? If so just wondering how much value there is in the incremental approach vs big bang.

server/utils/userUtils.test.ts
Harriethw: ah that's a good point yeah, will take a look at your PR

Harriethw:
@patrickjfl I think the incremental approach gives us some value now - the LCA's can then see prison dashboard ASAP. It's going to take me at least a day or two to give them write access as well. It's also more work to also remove the POM role (and that could potentially be a big change so we might want to get a bit more data on our users and be prepared to roll back when we remove that).

patrickjfl:
> @patrickjfl I think the incremental approach gives us some value now - the LCA's can then see prison dashboard ASAP. It's going to take me at least a day or two to give them write access as well (because it needs changes to API). It's also more work to also remove the POM role (and that could potentially be a big change so we might want to get a bit more data on our users and be prepared to roll back when we remove that).

Ah ok - I was under the impression that everyone with role LCA also has role POM, in which case they'd all be able to see the prison dashboard anyway, but is that not the case?

Harriethw:
> Ah ok - I was under the impression that everyone with role LCA also has role POM, in which case they'd all be able to see the prison dashboard anyway, but is that not the case?

I don't think there's any actual hard data for this! It's possible an LCA might not have role POM, according to an email and technically.
Having said that the API work has been a lot less than I expected to might just do all the UI in one go after all :)

Harriethw:
> Having said that the API work has been a lot less than I expected to might just do all the UI in one go after all :)
actually the API has gotta go through review etc. so might as well merge this as it's ready

patrickjfl:
> > Having said that the API work has been a lot less than I expected to might just do all the UI in one go after all :)
> 
> actually the API has gotta go through review etc. so might as well merge this as it's ready

👍

PR #658, opened 2024-04-24
Fix bug with HDC and conditional release date errors

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=CAS2-421

# Changes in this PR

Previously, invalid dates were being passed into isMoreThanMonthsBetweenDates, which was causing errors. This introduces return statements to ensure that empty or invalid dates are returned before those checks.

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/form-pages/apply/before-you-start/hdc-licence-dates/hdcLicenceDates.ts
+++ b/server/form-pages/apply/before-you-start/hdc-licence-dates/hdcLicenceDates.ts
@@ -12,6 +12,7 @@ import {
   isBeforeDate,
   differenceInDaysFromToday,
   isMoreThanMonthsBetweenDates,
+  dateIsComplete,
 } from '../../../../utils/dateUtils'
 import { dateBodyProperties } from '../../../utils'
 
@@ -76,10 +77,12 @@ export default class HDCLicenceDates implements TaskListPage {
 
   errors() {
     const errors: TaskListErrors<this> = {}
-    if (!dateAndTimeInputsAreValidDates(this.body, 'hdcEligibilityDate')) {
+
+    if (dateIsComplete(this.body, 'hdcEligibilityDate')) {
       errors.hdcEligibilityDate = "Enter the applicant's HDC eligibility date"
     }
-    if (!dateAndTimeInputsAreValidDates(this.body, 'conditionalReleaseDate')) {
+
+    if (dateIsComplete(this.body, 'conditionalReleaseDate')) {
       errors.conditionalReleaseDate = "Enter the applicant's conditional release date"
     }
     if (!dateIsTodayOrInTheFuture(this.body, 'conditionalReleaseDate')) {
--- a/server/form-pages/apply/before-you-start/hdc-licence-dates/hdcLicenceDates.ts
+++ b/server/form-pages/apply/before-you-start/hdc-licence-dates/hdcLicenceDates.ts
@@ -78,16 +78,24 @@ export default class HDCLicenceDates implements TaskListPage {
   errors() {
     const errors: TaskListErrors<this> = {}
 
-    if (dateIsComplete(this.body, 'hdcEligibilityDate')) {
+    if (!dateIsComplete(this.body, 'hdcEligibilityDate')) {
       errors.hdcEligibilityDate = "Enter the applicant's HDC eligibility date"
+    } else if (!dateAndTimeInputsAreValidDates(this.body, 'hdcEligibilityDate')) {
+      errors.hdcEligibilityDate = 'Eligibility date must be a real date'
     }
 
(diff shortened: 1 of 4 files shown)


server/utils/dateUtils.test.ts
```diff
@@ -375,4 +376,39 @@ describe('DateFormats', () => {
...
+      const date: ObjectWithDateParts<'field'> = {
+        'field-day': '12',
+        'field-month': '1',
+        'field-year': '2022',
+      }
+
+      expect(dateIsBlank(date, 'field')).toEqual(false)
+    })
+
+    it('returns true if the date is blank', () => {
```
tahb, line 391: 👋 Would it make sense to also cover the case where only some inputs were provided?

server/utils/dateUtils.ts
```diff
@@ -269,3 +269,10 @@ export const getTodaysDatePlusMonthsAndDays = (monthsToAdd = 0, daysToAdd = 0):
 }
 
 export class InvalidDateStringError extends Error {}
+
+export const dateIsBlank = <K extends string | number>(
+  dateInputObj: Partial<ObjectWithDateParts<K>>,
+  key: K,
+): boolean => {
+  return !['year' as const, 'month' as const, 'day' as const].every(part => !!dateInputObj[`${key}-${part}`])
```
tahb, line 277: Suggestion only: There's quite a bit happening here on a single line, especially when using `!`s to invert things.

I'd personally lean towards something more verbose for more readability, something like:

```
const dateParts = ['year', 'month', 'day'] as const;
let allPartsExist = true;

for (const part of dateParts) {
    const keyExists = Boolean(dateInputObj[`${key}-${part}`]);
    if (!keyExists) {
        allPartsExist = false;
        break;
    }
}
```

server/form-pages/apply/before-you-start/hdc-licence-dates/hdcLicenceDates.ts
```diff
@@ -76,15 +77,33 @@ export default class HDCLicenceDates implements TaskListPage {
...
+
+    if (dateIsBlank(this.body, 'hdcEligibilityDate')) {
       errors.hdcEligibilityDate = "Enter the applicant's HDC eligibility date"
     }
-    if (!dateAndTimeInputsAreValidDates(this.body, 'conditionalReleaseDate')) {
+    if (dateIsBlank(this.body, 'conditionalReleaseDate')) {
       errors.conditionalReleaseDate = "Enter the applicant's conditional release date"
     }
+
+    if (errors.hdcEligibilityDate || errors.conditionalReleaseDate) {
```
Harriethw, line 88: do we want to return early at this point? This would mean if one field was blank, but the other had an invalid date, it wouldn't capture it 
Screenshot 2024-04-24 at 16 48 49
would it make more sense to group the statements by field rather than condition? Something like
```
    if (dateIsBlank(this.body, 'hdcEligibilityDate')) {
      errors.hdcEligibilityDate = "Enter the applicant's HDC eligibility date"
    } else if (!dateAndTimeInputsAreValidDates(this.body, 'hdcEligibilityDate')) {
      errors.hdcEligibilityDate = 'Eligibility date must be a real date'
    }

    if (dateIsBlank(this.body, 'conditionalReleaseDate')) {
      errors.conditionalReleaseDate = "Enter the applicant's conditional release date"
    } else if (!dateAndTimeInputsAreValidDates(this.body, 'conditionalReleaseDate')) {
      errors.conditionalReleaseDate = 'Conditional release date must be a real date'
    } else if (!dateIsTodayOrInTheFuture(this.body, 'conditionalReleaseDate')) {
      errors.conditionalReleaseDate = 'Conditional release date cannot be in the past'
    }

    if (errors.hdcEligibilityDate || errors.conditionalReleaseDate) {
      return errors
    }
```
Is that really gross to have too many `else if`s I don't know it's the end of the day 😅  but I guess that is what we want to say here? We want to go through certain checks before getting to the next one...

tahb (dismissed):
Looks good! Some minor suggestions, I think we probably ought to add another small unit test and you can take or leave the thought on code style.

server/utils/dateUtils.ts
patrickjfl: yes, agree this is clearer. I've amended it 👍

server/utils/dateUtils.test.ts
patrickjfl: 👍

Harriethw (approved):
👍 👍 👍

PR #694, opened 2024-05-10
Fix content display on review application pages

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=CAS2-464

# Changes in this PR

On smaller screen displays, this content was layering, making it impossible to read. This media query fixes that.

However, tabbing doesn't provide a very satisfactory experience, because when you hit return on a specific item, you navigate there, but lose your place in the side nav component (so then need to tab all the way through the page to get back to it). This is a preexisting issue and I'm not sure how easy it would be to solve, but if anyone has any ideas, do share them.

### Before

Screenshot 2024-05-10 at 15 14 55

### After

Screenshot 2024-05-10 at 15 14 21

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/assets/scss/components/_side-nav.scss
+++ b/assets/scss/components/_side-nav.scss
@@ -1,4 +1,6 @@
 .side-nav {
-  position: sticky;
-  top: 0.5rem;
+  @include govuk-media-query($from: tablet) {
+    position: sticky;
+    top: 0.5rem;
+  }
 }


tahb:
A media query was my first instinct. Another way might be to look at what our layout system can do for us and if we're missing something? I imagine it might be something inside the [GOV.UK Design system layouts](https://design-system.service.gov.uk/styles/layout/). Doing something here might result in leaning on their media queries instead which should give us a more consistent experience.

patrickjfl:
> A media query was my first instinct. Another way might be to look at what our layout system can do for us and if we're missing something? I imagine it might be something inside the [GOV.UK Design system layouts](https://design-system.service.gov.uk/styles/layout/). Doing something here might result in leaning on their media queries instead which should give us a more consistent experience.

yes, although presumably users will want to navigate into the items on the side nav, so I'm not sure there's a straightforward solution. If they tab backwards then they are taken straight back to the side nav component from viewing the answers in the submitted application views (although in the CYA view they have to back tab through all the 'Change' links first).

Harriethw:
> A media query was my first instinct. Another way might be to look at what our layout system can do for us and if we're missing something? I imagine it might be something inside the [GOV.UK Design system layouts](https://design-system.service.gov.uk/styles/layout/).

@tahb are you suggesting this for fixing the position of the side nav rather than the tabbing issue?
I think we are using the gov uk layouts correctly here - they flex correctly, as in this screenshot, but the issue is that the side nav is 'sticky' so when the user scrolls the rest of the page appears underneath it. We want it to be sticky on desktop but don't want it to be sticky on mobile or tablet so we I think we need to do a media query.
Screenshot 2024-03-15 at 15 26 22

Harriethw:
> However, tabbing doesn't provide a very satisfactory experience, because when you hit return on a specific item, you navigate there, but lose your place in the side nav component

agree this is a bit annoying, but as you say it's pre-existing so would be worth doing in a separate ticket?
The only thing I could think is a script to re-focus the link so the keyboard focus is still on that link, but I'm not sure that would be a better user experience as for submitted applications they might want to click within the summary card...
It is possible to navigate back in tab order by pressing 'shift +tab' - doing that I was able to go back to the 'back to top' link and then tab through the content links.

Harriethw (approved):

PR #700, opened 2024-05-14
update 'no Oasys' nomsNumber

Harriethw:
Using work merged in this AP Tools PR: https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/67

We will shortly be restricting referrer
permissions to only create applications for
offenders within the same prison. The previous
nomsNumber used here did not have prison data, and a new one has been added in AP Tools with more
detail here:
https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/67s

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -70,7 +70,7 @@ export default defineConfig<TestOptions>({
         personWithoutOasys: {
           name: 'James Brown',
           crn: 'C246139',
-          nomsNumber: 'A5671YZ',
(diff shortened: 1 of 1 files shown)


Harriethw:
it looks like our Pact contract testing is failing on some types... will update in another PR

Harriethw:
oh there is an error in the shared types, following up here https://mojdt.slack.com/archives/C03K0HB0LBE/p1715770815911229?thread_ts=1715770716.112979&cid=C03K0HB0LBE

PR #701, opened 2024-05-14
Display nationality as "Unknown" when it is not present in person definition

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/CAS2-236

# Changes in this PR

As above. Arguably there are other person fields that we should do the same with, but I've limited this to nationality based on what has been flagged.

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/utils/checkYourAnswersUtils.test.ts
+++ b/server/utils/checkYourAnswersUtils.test.ts
@@ -6,6 +6,7 @@ import { formatLines } from './viewUtils'
 import applicationData from '../../integration_tests/fixtures/applicationData.json'
 import Apply from '../form-pages/apply'
 import { UnknownPageError } from './errors'
+import { DateFormats } from './dateUtils'
 
 jest.mock('./formUtils')
 jest.mock('./viewUtils')
@@ -20,6 +21,7 @@ const {
   getSections,
   getPage,
   getPages,
+  getApplicantDetails,
 } = checkYourAnswersUtils
 
 const { getQuestions } = getQuestionsUtil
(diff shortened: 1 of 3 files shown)


server/utils/checkYourAnswersUtils.ts
```diff
@@ -243,7 +243,7 @@ export const getApplicantDetails = (application: Application | Cas2SubmittedAppl
         text: 'Nationality',
       },
       value: {
-        html: nationality,
+        html: nationality || 'Unknown',
```
tahb, line 246: Should we add a unit test for this? I went looking but couldn't find this `getApplicantDetails` method tested at all in `checkYourAnswersUtils.test.ts`, so I suspect we may be able to create a prior commit that adds tests for the previous behaviour?

server/views/people/partials/personDetails.njk
```diff
@@ -22,7 +22,7 @@
         text: "Nationality"
       },
       value: {
-        text: person.nationality
+        text: person.nationality or 'Unknown'
```
Harriethw, line 25: do we need this here if 'Unknown' has already been set in the javascript function?

patrickjfl: yes, this one is to populate the 'Confirm applicant details' page, the other one is to populate CYA and submitted application views

server/utils/checkYourAnswersUtils.test.ts
```diff
@@ -517,4 +519,81 @@ describe('getPage', () => {
       getPage('confirm-eligibility', 'bar')
     }).toThrow(UnknownPageError)
   })
+
+  describe('getApplicantDetails', () => {
```
tahb, line 523: Now we have some test coverage I was thinking we could add a new shorter test for checking the default behaviour

```
person = personFactory.build({nationality: nil})
expected = {
          key: {
            text: 'Nationality',
          },
          value: {
            html: "Unknown",
          },
        },
 expect(getApplicantDetails(application)).toContain(expected)
```

patrickjfl: we could, but how much value is there in that test? Are we not at that point basically testing the javascript `||` operator?

tahb: The implementation is simple in this case but so is the test and having a test would protect it from regression and document the behaviour.

patrickjfl: OK yeah, good points. I have done this 👍

tahb (approved):

PR #666, opened 2024-04-26
Paginate applications on the prison dashboard page

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-409

# Changes in this PR

Adds pagination to the prison dashboard view.

### Before

Screenshot 2024-04-26 at 15 23 00

### After

PrisonDashboard -- sees tab for my applications

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Are any changes required to the e2e tests?
- [ ] If you've added a new route, have you added a new
  `auditEvent`? (see `server/routes/apply.ts` for examples)
- [ ] Are there environment variables or other infrastructure configuration which needs to be included in this release?
- [ ] Are there any data migrations required. Automatic or manual?
- [ ] Does this rely on changes being deployed to the CAS API?

## Post merge checklist

Once we've merged it will be auto-deployed to the dev environment.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-community-accommodation-tier-2-ui).

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #705, opened 2024-05-15
Fix e2e test for adding note via prison dashboard

Harriethw:
We need to test that our POM and LCA users can view and
add notes to applications on their prison dashboard.

Previously we hardcoded the application ID to be used locally
and in dev environment, because we had a standard
prison for our users across both environments.

Since we now are restricting users to only create applications
for offenders within their own prison, we have discovered that
the example offenders in dev and local have different prisons (LEI for local,
MDI for dev). [1] So our local and dev test users do not have the same prison
any more, and that hardcoded UUID no longer works.

In this commit we use the user's name to find an application by any user
that is not them, thus removing the need for a hardcoded id.

This has the added benefit that it's agnostic to what the seeded data is,
and when [we introduce pagination we don't care what page we are on.](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/666)

[1] https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1879

### DIFF
--- a/e2e-tests/steps/apply.ts
+++ b/e2e-tests/steps/apply.ts
@@ -117,8 +117,18 @@ export const addNote = async (page: Page) => {
   await expect(page.locator('.moj-timeline__title').first()).toContainText('Note')
 }
 
-export const viewApplicationFromPrisonDashboard = async (applicationId: string, page: Page) => {
+export const goToPrisonDashboard = async (page: Page) => {
   await page.goto(`/applications/prison`)
-  await page.locator(`a[href="/applications/${applicationId}/overview"]`).click()
+}
+
+export const checkAnApplicationByUserExists = async (page: Page, name: string) => {
+  const tableRows = page.locator('.govuk-table__row')
+  expect(tableRows.filter({ hasText: name }).first()).toBeVisible()
+}
+
+export const viewApplicationMadeByAnotherUser = async (page: Page, name: string) => {
+  const tableRows = page.locator('.govuk-table__row')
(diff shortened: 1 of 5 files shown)


e2e-tests/steps/apply.ts
```diff
@@ -117,8 +117,12 @@ export const addNote = async (page: Page) => {
...
 }
 
-export const viewApplicationFromPrisonDashboard = async (applicationId: string, page: Page) => {
+export const goToPrisonDashboard = async (page: Page) => {
   await page.goto(`/applications/prison`)
-  await page.locator(`a[href="/applications/${applicationId}/overview"]`).click()
+}
+
+export const viewApplicationMadeByAnotherUser = async (page: Page, name: string) => {
+  const tableRow = page.locator('.govuk-table__row').filter({ hasNotText: name }).last()
```
tahb, line 125: Would it give us better test feedback if we first had another expectation which said "expect table to include the authors name"?

tahb (approved):
LGTM

e2e-tests/steps/apply.ts
Harriethw: that's a good idea - that should work if it's always run after the 'create an application' test, which it should do on CI.  The only time it might be flakey is running it locally, if we are running the tests out of order there may not be a recent application made by the user, but I think the trade off is we are getting more coverage on CI.

Harriethw: done

PR #713, opened 2024-05-21
Update govuk-frontend to v5.4.0

patrickjfl:
# Context

# Changes in this PR

### After

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/assets/scss/application.scss
+++ b/assets/scss/application.scss
@@ -4,7 +4,12 @@ $path: '/assets/images/';
 $moj-page-width: 1170px;
 $govuk-page-width: $moj-page-width;
 
-@import 'govuk/all';
+@import 'govuk/base';
+@import 'govuk/core/index';
+@import 'govuk/objects/index';
+@import 'govuk/components/index';
+@import 'govuk/utilities/index';
+@import 'govuk/overrides/index';
 @import 'moj/all';
 
 @import './components/header-bar';
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "express-prom-bundle": "^7.0.0",
         "express-session": "^1.17.3",
         "fishery": "^2.2.2",
-        "govuk-frontend": "^5.0.0",
+        "govuk-frontend": "^5.4.0",
         "helmet": "^7.0.0",
         "http-errors": "^2.0.0",
         "husky": "^8.0.3",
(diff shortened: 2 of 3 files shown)


assets/scss/application.scss
```diff
@@ -4,7 +4,7 @@ $path: '/assets/images/';
 $moj-page-width: 1170px;
 $govuk-page-width: $moj-page-width;
 
-@import 'govuk/all';
+@import 'govuk/index';
```
tahb, line 7: It can be helpful to link directly to the release notes https://github.com/alphagov/govuk-frontend/releases/tag/v5.4.0 just to keep everything together and make it super simple for the reviewer.

assets/scss/application.scss
```diff
@@ -30,7 +30,7 @@ $govuk-page-width: $moj-page-width;
 }
 
 .govuk-panel--warning {
-  background-color: #FAC710;
+  background-color: #fac710;
```
tahb, line 33: A very minor change obviously but I don't _think_ this is needed as part of the upgrade? Guessing it was a style thing you noticed in passing? I'd put this in another commit so it's separate from the riskier change should that part need to be reverted for whatever reason, you're other work will persist 👍

assets/scss/application.scss
tahb: Did this work for you locally? I got an error on this line where index couldn't be found. When I changed it to the following (based on the release notes) it worked:

```
@import "govuk/base";
@import "govuk/core/index";
@import "govuk/objects/index";
@import "govuk/components/index";
@import "govuk/utilities/index";
@import "govuk/overrides/index";
```

assets/scss/application.scss
patrickjfl: hmmm very strange, I didn't change this at all. It must have happened automatically somehow. Anyway I will remove it, thanks

assets/scss/application.scss
patrickjfl: it did work for me locally, but let me try again.

patrickjfl: OK now it's not. I thought I had deleted the stylesheets previously to reset the thing but I guess not. I will update this with your suggestion (which also works locally for me).

tahb (approved):
LGTM 👍

PR #727, opened 2024-05-28
Cas2-216: Cancel from dashboard

Harriethw:
~~NOT TO BE MERGED until API PR has been https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1917~~ ✅ 

# Context

https://dsdmoj.atlassian.net/browse/CAS2-216

As a referrer
I want to have the ability to withdraw an (in-progress) application
So that I can ensure my dashboard remains tidy and clear of any applications started in error

# Changes in this PR

This introduces a new Controller to handle the getting and posting of the form view to cancel.
In UI-land we are calling this functionality 'Cancel' based on previous UCD work - in the API we are calling this 'Abandon' because we want to make it clear that it only applies to in progress applications, and distinguish it from the 'Referral cancelled' Status Update. 

### After
Screenshot 2024-05-28 at 11 56 16
Screenshot 2024-05-28 at 09 45 19

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/e2e-tests/steps/cancelInProgressApplication.ts
+++ b/e2e-tests/steps/cancelInProgressApplication.ts
@@ -1,0 +1,12 @@
+import { Page } from '@playwright/test'
+import { ApplyPage } from 'e2e-tests/pages/apply'
+
+export const cancelAnApplication = async (page: Page, name: string) => {
+  const cancelPage = await ApplyPage.initialize(page, `Are you sure you would like to cancel ${name}'s application?`)
+  await cancelPage.checkRadio('Yes, cancel the application')
+  await cancelPage.clickButton('Confirm')
+}
+
+export const clickCancel = async (page: Page) => {
+  await page.getByRole('link', { name: 'Cancel' }).last().click()
+}
--- a/server/views/applications/index.njk
+++ b/server/views/applications/index.njk
@@ -5,6 +5,7 @@
 {% from "../partials/showErrorSummary.njk" import showErrorSummary %}
 {% from "../components/formFields/form-page-input/macro.njk" import formPageInput %}
 {% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
 {% from "../partials/table.njk" import applicationsTable %}
 
 {% extends "../partials/layout.njk" %}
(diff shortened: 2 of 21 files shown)


server/controllers/apply/applicationsController.ts
```diff
@@ -302,4 +302,46 @@ export default class ApplicationsController {
       })
     }
   }
+
+  cancelForm(): RequestHandler {
```
Harriethw, line 306: I wasn't sure how best to do these controller functions. I need one to render the form view, and the other one to be the endpoint the form POSTs to 🤔

tahb: We could leave those methods here or extract them into a new  abandon controller? From there we can continue to use conventional naming such as #new #create methods.

Perhaps similar to how we have the PagesController nested under controllers/apply/applications/pagesController.ts

server/views/applications/index.njk
```diff
@@ -62,7 +62,8 @@
                   },
                   {
                     text: "Date started"
-                  }
+                  },
+                  {}
```
tahb, line 66: Is there a way we can add some text here that doesn't render but explains what it is to screen readers?

I'm guessing we don't have a way of sending in the following?

```
<div aria-label="test A"></div>
```

I wonder what the axe plugin says but I wonder if having the text "Actions" would be reasonable.

e2e-tests/steps/cancelInProgressApplication.ts
```diff
@@ -0,0 +1,12 @@
+import { Page } from '@playwright/test'
+import { ApplyPage } from 'e2e-tests/pages/apply'
+
+export const cancelAnApplication = async (page: Page, name: string) => {
+  const cancelPage = await ApplyPage.initialize(page, `Are you sure you would like to cancel ${name}'s application?`)
```
tahb, line 5: Are we reusing the Page object here just to help with testing? I wouldn't have thought to do that, very nice!

tahb (approved):
LGTM!

The frontends use of the 'cancel' terminology for more than just the view gave me pause given the choice of 'abandon' on the API side. I had thought it would make sense to make them match so devs can see relate on both sides. This would leave the language of 'cancel' to be presented to users and tested where renaming it is decoupled from the rest of the code base. Saying that I don't think it's worth renaming now given it's more imp to get this released before we leave 👍

server/views/applications/index.njk
Harriethw: ah I see what you mean, will take a quick look

Harriethw: So the govuk component has the ability to [add `html` within the `th` element](https://design-system.service.gov.uk/components/table/#options-table-example--head), but not to add a role or label to the actual `th` element. I think from [looking at accessibility guidelines](https://github.com/dequelabs/axe-core/issues/3404) that if we have a tableheader/columnheader then it should have some kind of name to it as thats semantically what it is there for... I'm going to err on the side of having a visible 'Actions' text within the `th` as it may not be exactly what the visual design was but I think it will make it more accessible.

PR #731, opened 2024-05-29
Chore/check your answers uses safe page keys

tahb:
This is a proposal to make removing a task from the application more robust. We've run out of time to release this and so need to hand it over.

# Context

[We experienced a production error after removing removing a section from the question form](https://mojdt.slack.com/archives/C05FN6Y2CSV/p1716972264605899). They already had a referral in progress when we released a change to remove a task from the journey, they were then unable to render their check your answers page. We released a fix the user was able to get their existing in progress referral through to submission. 

This would have affected every in progress referral so the potential impact of this bug was much more than 1 user.

This case highlights a fragility in the question framework. When we want to remove a task we have to remember [to hard code that page key into a list of questions to avoid](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/60554b22eec1f11d2735d551d95356156bf11415/server/utils/checkYourAnswersUtils.ts#L44). This is easily forgotten as this case demonstrates.

```typescript
pagesKeys.forEach(pageKey => {
    if (!['behaviour-notes', 'behaviour-notes-data', 'reducing-risk'].includes(pageKey)) {
      addPageAnswersToItemsArray({
        items,
        application,
        task,
        pageKey,
        questions,
        outputFormat,
      })
    }
  })
```

N.B when a _question_ is removed, as opposed to a _task_ [this logic is currently protecting us dynamically](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/60554b22eec1f11d2735d551d95356156bf11415/server/utils/checkYourAnswersUtils.ts#L93).

As a way to improve this we had the idea to dynamically check the latest tasks against the content of the application. When we find a tasks that were present on creation but are no longer supporter we automatically ignore them.

The remaining tasks are to make sure the unit tests for checkYourAnswers are updated to cover this change in design.

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/server/utils/checkYourAnswersUtils.test.ts
+++ b/server/utils/checkYourAnswersUtils.test.ts
@@ -767,19 +767,27 @@ describe('getPage', () => {
   })
 
   describe('removeAnyOldPageKeys', () => {
-    it('should remove any task keys that appear in the application but are no longer part of the latest set of questions', () => {
+    it('should remove any page keys that appear in the application but are no longer part of the latest set of questions', () => {
       const questions = {
         task1: {
           page1: { question1: { question: 'A question', answers: { yes: 'Yes', no: 'No' } } },
           page2: { question2: { question: 'Another question' } },
         },
       } as unknown as Questions
-
       const applicationPageKeys = ['page1', 'page2', 'oldPageKey']
-
       const expected = ['page1', 'page2']
-
       expect(removeAnyOldPageKeys(questions, 'task1', applicationPageKeys)).toEqual(expected)
     })
+  })
+  it('should retain ACCT, current offences and historic offences even where they do not appear in the question schema', () => {
+    const questions = {
+      task1: {
+        page1: { question1: { question: 'A question', answers: { yes: 'Yes', no: 'No' } } },
+        page2: { question2: { question: 'Another question' } },
+      },
+    } as unknown as Questions
+    const applicationPageKeys = ['page1', 'page2', 'oldPageKey', 'acct', 'current-offences', 'offence-history']
+    const expected = ['page1', 'page2', 'acct', 'current-offences', 'offence-history']
+    expect(removeAnyOldPageKeys(questions, 'task1', applicationPageKeys)).toEqual(expected)
   })
 })
--- a/server/utils/checkYourAnswersUtils.test.ts
+++ b/server/utils/checkYourAnswersUtils.test.ts
@@ -22,6 +22,7 @@ const {
   getPage,
   getKeysForPages,
   getApplicantDetails,
+  removeAnyOldPageKeys,
 } = checkYourAnswersUtils
 
 const { getQuestions } = getQuestionsUtil
@@ -78,10 +79,6 @@ describe('checkYourAnswersUtils', () => {
 
   describe('getTaskAnswersAsSummaryListItems', () => {
     it('returns an array of summary list items for a given task', () => {
-      jest.spyOn(getQuestionsUtil, 'getQuestions').mockImplementationOnce(jest.fn(() => mockQuestions))
-
-      jest.spyOn(checkYourAnswersUtils, 'getPage').mockReturnValue(jest.fn())
-
       const mockApplication = applicationFactory.build({
         data: {
           task1: {
@@ -94,6 +91,10 @@ describe('checkYourAnswersUtils', () => {
           },
         },
       })
+
+      jest.spyOn(getQuestionsUtil, 'getQuestions').mockImplementationOnce(jest.fn(() => mockQuestions))
+
+      jest.spyOn(checkYourAnswersUtils, 'getPage').mockReturnValue(jest.fn())
 
       const expected = [
         {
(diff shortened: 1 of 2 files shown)


server/utils/checkYourAnswersUtils.ts
```diff
@@ -293,3 +298,33 @@ export const getApplicantDetails = (application: Application | Cas2SubmittedAppl
...
+const processFindMatchedKeys = async (jsonBlob: any, keysToFind: string[]): Promise<string[]> => {
+  const keys = Object.keys(jsonBlob)
+  const result = await Promise.all(keys.map(key => findMatchedKeys(jsonBlob[key], keysToFind)))
+  return result.flat()
+}
+
+const filterKeysWithoutQuestions = async (questions: any, task: string, pageKeys: string[]): string[] => {
+  const permanentPagesWithoutQuestions = ['summary-data', 'oasys-import']
+  const dynamicPagesKeys = pageKeys.filter(item => !permanentPagesWithoutQuestions.includes(item))
+  const matchedKeys = await processFindMatchedKeys(questions[task], dynamicPagesKeys)
```
tahb, line 327: We fan out across the list of keys here and each task will drill down and compare keys. Rather than going through each key and sub key sequentially.

server/utils/checkYourAnswersUtils.test.ts
```diff
@@ -78,9 +79,12 @@ describe('checkYourAnswersUtils', () => {
 
   describe('getTaskAnswersAsSummaryListItems', () => {
     it('returns an array of summary list items for a given task', () => {
-      jest.spyOn(getQuestionsUtil, 'getQuestions').mockImplementationOnce(jest.fn(() => mockQuestions))
-
-      jest.spyOn(checkYourAnswersUtils, 'getPage').mockReturnValue(jest.fn())
+      const mockQuestionsForOneTask = {
```
tahb, line 82: We redefine the mock questions with just one task, since that's all we care about in this scenario.

server/utils/checkYourAnswersUtils.ts
```diff
@@ -36,23 +36,26 @@ export const getTaskAnswersAsSummaryListItems = (
 ): Array<SummaryListItem | QuestionAndAnswer> => {
   const items: Array<SummaryListItem | QuestionAndAnswer> = []
 
+  // Get the latest question schema
```
tahb, line 39: I try not to add comments but this code felt like variable and method naming alone didn't provide enough.

tahb:
Looks like I've got a failing integration test at the last minute so will still need to hand this over with a tiny bit more to do @froddd @patrickjfl.

The failure is where it can't find "Arson". Presumably the new code is filtering this out as is viewing it as old question.

Screenshot 2024-05-31 at 17 46 36

https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/b05f338a698f290f3c1974f3bfd40c5b9d0bb1f3/integration_tests/tests/apply/check_your_answers/check_your_answers.cy.ts#L19 points to the applicationData.json [where "arson" appears in both](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/b05f338a698f290f3c1974f3bfd40c5b9d0bb1f3/integration_tests/fixtures/applicationData.json#L303) "current-offences" and "offending-history". [The current questions.ts says that arson is not a type of current offence?](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/b05f338a698f290f3c1974f3bfd40c5b9d0bb1f3/server/form-pages/utils/questions.ts#L795), is the fix here to remove it from the applicationData.json?

server/utils/checkYourAnswersUtils.test.ts
froddd: Not entirely sure this makes this test more readable, having the same questions context for the whole test suite seems fairly easy to grok -- effectively this only duplicates a subset of the existing context test data?

froddd (approved):
Seems sensible and does the job it needs to do, I may feel a slight lack of linear readability but this may well be due to not being familiar with this codebase!

server/utils/checkYourAnswersUtils.test.ts
patrickjfl: yes, avoiding a long and verbose expectation statement. But happy to defer if you want to set it up differently @froddd

server/utils/checkYourAnswersUtils.ts
```diff
@@ -36,23 +36,26 @@ export const getTaskAnswersAsSummaryListItems = (
...
-        pageKey,
-        questions,
-        outputFormat,
-      })
-    }
+  // Get the page keys stored on the application at creation
+  const applicationPageKeys = getKeysForPages(application, task).filter(key => pagesWeNeverWantToPresent(key) === false)
+
+  // Filter out any keys that are no longer in the latest question schema
+  const relevantPagesKeys = removeAnyOldPageKeys(questions, task, applicationPageKeys)
```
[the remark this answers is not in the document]
patrickjfl: Personally I think it makes more sense to have two functions here, for readability and separation of concerns. We want getKeysForPages to go and get all the page keys for the task from the application data blob. And then we want to go and verify these keys, against an up to date list of the keys that we need. I think doing all that in one function probably makes things slightly harder to follow?

server/utils/checkYourAnswersUtils.test.ts
froddd: The expectation statement shouldn't change, given the method under test is called with `task1` specifically. Removing `mockQuestionsForOneTask` and passing `mockQuestions` to the tested method tests the same assertion and doesn't require the expectation to change, which is why I find it a bit odd to duplicate it in this case.

patrickjfl: ah I see what you mean! yes, sorry Fred, good point and I agree. I've removed it 👍

PR #745, opened 2024-06-10
CAS-401: CAS2 ip ranges updates

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-401

# Changes in this PR

CAS2 does not apply any allowlist to control access, instead relying on auth and MFA (https://dsdmoj.atlassian.net/wiki/x/R4G2DgE), so there is no need to add extra ranges. This PR moves the default `allowlists: null` setting to the base config, and removes confusing overrides per environment.

## Post merge checklist

fix it as soon as possible.

### DIFF
--- a/helm_deploy/hmpps-community-accommodation-tier-2-ui/values.yaml
+++ b/helm_deploy/hmpps-community-accommodation-tier-2-ui/values.yaml
@@ -56,9 +56,8 @@ generic-service:
       REDIS_HOST: "primary_endpoint_address"
       REDIS_AUTH_TOKEN: "auth_token"
 
-  allowlist:
-    groups:
-      - internal
+  # CAS2 relies on MFA for managing access and therefore does not require allowlists, see: https://dsdmoj.atlassian.net/wiki/x/R4G2DgE
+  allowlist: null
 
 generic-prometheus-alerts:
   targetApplication: hmpps-community-accommodation-tier-2-ui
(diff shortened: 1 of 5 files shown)


helm_deploy/hmpps-community-accommodation-tier-2-ui/values.yaml
```diff
@@ -56,9 +56,8 @@ generic-service:
       REDIS_HOST: "primary_endpoint_address"
       REDIS_AUTH_TOKEN: "auth_token"
 
-  allowlist:
-    groups:
-      - internal
```
edavey, line 61: hmm, I think we probably *do* want to apply IP filtering to all environments other than production? i.e. it makes sense for our internal development and qa environments to be limited to that `internal` group?

froddd: I had the same thought but didn't want to deviate from current implementation, which actually sets `null` on **all** environments. I think it would make sense, to have the internal group for all but prod. If there's consensus I'm happy to update this, does this need confirming that there are no need for users from `prison` group to access dev/test/pre-prod? I.e. for user research, debugging, etc?

edavey: I think `internal` will cover all CAS2 users of dev/test/pre-prod but you should maybe check with the product owner? If, for example, certain prison staff needed to use pre-prod then you could add the `prisons` group from `hmpps-ip-allowlists` to the preprod helm values. But if NACRO needed to use pre-prod then you'd need to remove all IP controls from pre-prod. But I don't believe that's currently the case.

froddd:
Given no change is needed for the extra IP ranges that have been requested, I'm closing this until we get clarification on whether we should apply allowlists to lower envs, as this would be a separate concern.

PR #769, opened 2024-07-04
APS-999 Use #cas-events for all circle slack alerts

davidatkinsuk:


froddd:
This is now superseded by #772

PR #788, opened 2024-07-24
Retain trace and video on E2E test failure

richpjames:
If the test fails in CI we want to be able to playback the trace and/or video in order to better understand what went wrong.


davidatkinsuk (approved):
LGTM 👍

PR #795, opened 2024-08-01
CHORE: Address fast-xml-parser vulnerability

froddd:


williamfalconeruk (commented):
was there an update to the package.json to go with this?

froddd:
@williamfalconeruk no, typically npm package vulnerabilities are fixed by pinning a more recent version in the lock file -- they're often indirect dependencies so this is sufficient.

froddd:
Closed in favour of Dependabot PR #794

PR #797, opened 2024-08-12
CAS2-505 Change filter note from 14 to 32 days.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-505

Applications where the status is withdrawn, cancelled or awaiting arrival and created greater than 32 days in the past can be excluded from the applications list (previously 14 days).

Related API PR: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2173


davidatkinsuk (approved):
LGTM 👍

PR #802, opened 2024-08-17
improve "Remove" links

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS2-452

"remove was being read out twice for screen reader users because the visual text is "Remove" and then  repeated again in the visually hidden text. Although not mentioned in the audit report screenreader users would struggle to navigate the links if there were multiple links with the same text but taking them to different pages. I've tried to improve it by giving more context to the Remove link in the visually hidden text and will double check with content design

No visual differences for the user.


froddd (approved):
👍

PR #809, opened 2024-09-19
Trim user input strings

aliuk2012:
Repeat of previous config added to CAS 1 here https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2137

### DIFF
--- a/server/middleware/trimInput.ts
+++ b/server/middleware/trimInput.ts
@@ -1,0 +1,17 @@
+import { RequestHandler } from 'express'
+
+const trimStringValues = (body?: Record<string, unknown>) =>
+  body
+    ? Object.entries(body).forEach(([key, value]) => {
+        body[key] = typeof value === 'string' ? value.trim() : value
+      })
+    : undefined
+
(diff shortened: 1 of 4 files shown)


server/middleware/trimInput.ts
```diff
@@ -0,0 +1,17 @@
+import { RequestHandler } from 'express'
+
+const trimStringValues = (body?: Record<string, unknown>) =>
```
bobmeredith, line 3: Not sure the body param needs to be optional.

froddd (approved):
👍

bobmeredith (approved):
Looks good

server/middleware/trimInput.ts
froddd: You might be right actually -- the Express docs specify that, [by default, `req.body` is `undefined`](https://expressjs.com/en/api.html#req.body), but logging both `req.body` and `req.query` on a page with no values expected does return `{}` for both. This might need to be updated, then!

PR #816, opened 2024-09-30
Remove Risk factors page from application process

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS2-364

# Changes in this PR
Removed two fields because we are caturing this information already from users.

- "What circumstances are likely to increase risk?"
- "When is risk likely to be greatest?"

POMs will not longer be required to fill in this information for new applications. Historical data will exist but will eventually be phased out. The reason for not removing/hiding existing data is because we know that this information will eventually be phased out and do not want to add conditional statements in the code to hide these two questions from users.

### Before
image

### After
image

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


integration_tests/pages/apply/risks_and_needs/risk-of-serious-harm/riskFactorsPage.ts
bobmeredith, line 1: Git is showing this file as empty rather than deleted

bobmeredith (approved):
LGTM - just one query

PR #845, opened 2024-12-23
CAS-948 Replace custom timeline with official moj design system component

aliuk2012:
# Context

https://dsdmoj.atlassian.net/browse/CAS-948

# Changes in this PR

Initially raised as an accessibility concern about document hierarchy, we reviewed our usage of the timeline component and decided to remove `Latest activity` from the first item as its not standard design in MoJ or CAS-1 & CAS-3. In doing this, its also meant that we no longer have to have to use custom timeline partial but can now using MoJ Frontend version directly.

I've run the E2E tests locally and everything still passes

### Before
image

### After
image


server/utils/dateUtils.ts
```diff
@@ -77,7 +77,7 @@ export class DateFormats {
    * @returns the date in the to be shown in the UI: "Thursday, 20 December 2012".
    */
   static isoDateTimeToUIDateTime(isoDate: string) {
-    return format(DateFormats.isoToDateObj(isoDate), "d MMMM y 'at' HH:mmaaa")
+    return format(DateFormats.isoToDateObj(isoDate), "d MMMM y 'at' h:mmaaa")
```
bobmeredith, line 80: Is this deliberate? 
This would be 12hr clock, but there is no AM/PM indicator in the format

server/utils/nunjucksSetup.ts
```diff
@@ -92,6 +95,12 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('documentSummaryListRows', documentSummaryListRows)
   njkEnv.addGlobal('prisonDashboardTableRows', prisonDashboardTableRows)
 
+  const mojFilters = getMojFilters()
+
+  Object.keys({ ...mojFilters, mojDate: DateFormats.isoDateTimeToUIDateTime }).forEach(filter => {
```
bobmeredith, line 100: Nice!

bobmeredith (approved):
Looks good.
I had one query about the switch of date format.

server/utils/dateUtils.ts
aliuk2012: See the before/after screenshots in the PR descriptions

PR #848, opened 2024-12-24
CAS-947 PDF title text does not meet contrast ratio requirements

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-947

# Changes in this PR

See individual commits but basically tried to tidy up "Print/Download as PDF" output by using a print stylesheet. 

### Before
image

### After
image

### Before of Downloaded PDF
image

### After of Downloaded PDF
image

### DIFF
--- a/assets/scss/application.scss
+++ b/assets/scss/application.scss
@@ -53,10 +53,6 @@ $govuk-page-width: $moj-page-width;
 }
 
 @media print {
-  .print__hidden {
-    display: none;
-  }
-
   .print__show {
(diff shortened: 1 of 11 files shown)


bobmeredith (approved):
LGTM.
A couple of things I noticed - 

- In the 'before' screengrab, there is an email address shown - but not in the 'after'. I don't think the PR changes were responsible however.
- Also, it might be better to remove the underline on the contact number in the PDF if it's no longer a link.

PR #854, opened 2024-12-30
CAS-950 Make the first nav link focusable for "back to top"

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-950

# Changes in this PR
In the sidebar, we have a long list of navigation items and the last one says "Back to top". Currently it sets the viewport to the top of page but does not focus any element. There is no anchor point with an id of `top` which is why the viewport takes the user to the top instead of top of the navigation list.

This commit ensures that the first link in the sidebar will now have an id of `top` which the `Back to top` can go to and focus the user on.
### Before

https://github.com/user-attachments/assets/38413c25-4dc5-434d-bc49-8948c0e4f9ba

### After

https://github.com/user-attachments/assets/5790929e-ab8c-4fa5-877f-8b4292ad81f6

# Release checklist

confirm:

## Pre merge checklist

## Post merge checklist

fix it as soon as possible.


muhammad-elabdulla (approved):
LGTM

PR #901, opened 2025-04-02
Add GH Actions for CI

aliuk2012:
# Context
This pipeline is based off CAS1 https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/f44b0f2481e1a5138b31d1bd6ba458090672263c/.github/workflows/pipeline.yml

This PR does not remove the  Circle CI config part of the workflow but I will do it shortly after to ensure the integration tests are being run in parallel correctly. For this check we need to have some previously run tests where the changes have been merged to main branch.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,0 +1,313 @@
+name: Pipeline [test -> build -> deploy]
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Environment
+        type: choice
+        required: true
+        options:
+          - test
+          - dev
+          - preprod
+          - prod
+        default: 'dev'
+      version:
+        description: Image version
+        type: string
+        required: true
+      docker_registry:
+        description: Docker registry
+        required: true
+        default: 'ghcr.io'
+        type: choice
+        options:
+          - 'ghcr.io'
+          - 'quay.io'
+      registry_org:
+        description: Docker registry organisation
+        required: true
+        default: 'ministryofjustice'
+        type: choice
+        options:
+          - 'ministryofjustice'
+          - 'hmpps'
+      additional_docker_tag:
+        description: Additional docker tag that can be used to specify stable tags
+        required: false
+        default: ''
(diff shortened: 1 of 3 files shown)


Rosie-Brigham (approved):
Looks good, thank you :)

froddd (commented):
I wonder if the test, build and deploy steps should be split into separate workflows? GH Actions is able to trigger another local workflow easily, and it means we would be able to run a deployment for a known build directly, without waiting for checks. Of course we could also specify that any deployment to prod _does_ run those tests... Everything is possible!

Another question: do we need this many different inputs for the workflow dispatch? Is this just to refine configuration while testing, then we can set the values that work?

One last thing: I've been wondering whether splitting the integration tests by actual test time is necessary. The splitting logic defaults to creating a matrix based on files, I think, and from what I can remember the gains from splitting using actual test times wasn't huge, making me question the overhead of having to 'fix' and save the XML output as artefact, and downloading them for each run... This also includes another external action, and I think minimising our reliance on those would be generally a good idea.

aliuk2012:
Thanks @froddd 

Really good points, I think its a case of getting something in and then refining it further. 

I agree, we can split them out into separate workflows and potentially create shared workflows across all the UI.  There are a lot of inputs here but only because I copied it over from CAS1 and looking to eventually do deployments

Regarding spliting integration tests, for CAS3 its not worth it at the moment because its super quick running the suite but for cas2, the integration tests are alot more sluggish so running them in parallel is a must. Again we can refine it further once we are all using the same thing

froddd:
Re: splitting the tests: I was talking about splitting _based on test timings_ specifically: I don't think there's much to gain from the steps `Downloading test reports from previous runs`, `Fix Cypress JUnit XML reports` and `Store Integration tests reports`. Without those reports, the splitting would still work fine, but would simply split by test name (I think?), which was actually producing very reasonable results for CAS1, compared to doing all of those steps to split by actual timings.

I'm not sure applying test pipeline configuration to CAS2 is right here, to be fair -- should we not settle on a working configuration on one of the CAS UIs, then apply it once we have something stable? It seems like we will now have a lot of different workflows to update, and they will possibly (likely) diverge 🫤

PR #902, opened 2025-04-02
Remove unit/integration tests, linting from CircleCI

aliuk2012:
Theres jobs are now run using GH Actions (https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/901)

It was discovered as part of this work that the splitting of the tests were not configured correctly and running all the tests 4 times. This was because we hadn't passed the TEST_ARG env var to the calling npm script. The 94 specs are now ~25 tests being run in each CI node.


davidatkinsuk (approved):
LGTM 👍

PR #908, opened 2025-04-08
Change bootstrap script to generate `.env` 

gregkhawkins:
# Context
* Generate .env file using new `.env.template` file
* Pull secrets down from k8s to swap out secret values in `.env`
* e2es now run against latest version of AP tools (that by default integrates with upstream dev services)


davidatkinsuk (commented):
Looks great!

PR #912, opened 2025-04-11
Integrate with new AP tools

gregkhawkins:
# Context
* As part of firebreak we have configured AP tools to run against dev-upstream services and removed running those services in our local docker infrastructure
* The changes in this PR enable running against this new version. of AP tools

# Changes in this PR
* Remove all local config for e2e tests as no longer required
* Change bootstrap script to generate `.env` file from new `.env.template` and resolve secret values using new `resolve_secrets.sh` script
* Change bootstrap script to generate `e2e.env` file from new `e2e.env.template` and resolve secret values using new `resolve_secrets.sh` script
* Change playwright config to use the new `e2e.env.template` to drive the user logins and e2e data


davidatkinsuk (approved):
LGTM

PR #924, opened 2025-04-29
Fix container registry path

froddd:
Container registry was updated to `ghcr.io` but path to organisation was still incorrect 🤦‍♂️ ... This now matches the path used for other CAS UIs:

- CAS1: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/helm_deploy/hmpps-approved-premises-ui/values.yaml#L8
- CAS3: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/helm_deploy/hmpps-temporary-accommodation-ui/values.yaml#L9


davidatkinsuk (approved):
LGTM 👍

PR #935, opened 2025-05-20
Chore/update log levels

froddd:
This PR sets the default log level to `info` for all environments. This can be overridden locally by setting a `LOG_LEVEL` environment variable equal to the desired level (`"trace" | "debug" | "info" | "warn" | "error" | "fatal"`).

The PR also removes any Personally Identifiable Information from any of the logging calls, instead showing the staff ID where relevant only.


gregkhawkins (approved):
LGTM

PR #950, opened 2025-06-27
Update generate-types job

daveawc:
this updates the generate types job to allow reading spec files from the dev environment. This is set by a repository parameter.

### DIFF
--- a/script/generate-types
+++ b/script/generate-types
@@ -6,9 +6,15 @@ set -e
 
 cd "$(dirname "$0")/.."
 
-echo "==> Generating types..."
+if [ -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


script/generate-types
```diff
@@ -6,9 +6,15 @@ set -e
 
 cd "$(dirname "$0")/.."
 
-echo "==> Generating types..."
+if [ -z "$1" ]; then
+  url="https://raw.githubusercontent.com/ministryofjustice/hmpps-approved-premises-api/main/src/main/resources/static/codegen/built-cas2-api-spec.yml"
```
davidatkinsuk, line 10: do we want to update these to be dev urls? I guess we'll need to do that once we stop using yml?

PR #952, opened 2025-07-01
Chore/cas 1807 update repo to read api spec in dev

daveawc:
removed use of OasysSection


server/testutils/factories/oasysSections.ts
```diff
@@ -42,14 +42,9 @@ export const offenceDetailsFactory = Factory.define<OASysQuestion>(options => ({
 }))
 
 export const supportingInformationFactory = Factory.define<OASysSupportingInformationQuestion>(options => {
```
davidatkinsuk, line 44: FYI i'm pretty sure this is dead code as CAS3 doesn't use the 'selection' endpoint (which i don't think even exists anymore).

PR #1070, opened 2026-04-15
Chore: Fix diff and serialize-javascript vulnerabilities

froddd:
This overrides the versions of `diff` and `serialize-javascript` to ensure no vulnerable version is used.


robanderson-moj (approved):
👍

PR #1074, opened 2026-04-15
Chore: Specify repository for token generation

froddd:
This ensures the token generated has permissions for the correct repository, even when generated from an external workflow (i.e. API).


danhumphreys-moj (approved):
Discussed with @froddd this afternoon.
This owner and repository are required so that the API pipeline has the necessary permission to raise a PR in the UI repository when the generate-types workflow is run within the context of the API deployment pipeline.  This dev token is ephemeral and revoked once the pipeline is run.

PR #1076, opened 2026-04-16
Chore: Fix follow-redirects vulnerability

froddd:
Dependabot unable to create PR, so here's a manual package-lock bump instead.


robanderson-moj (approved):
👍

PR #1078, opened 2026-04-23
Chore: Update Renovate config to match typescript template

froddd:
This newer config should help also keeping actions up to date


davidatkinsuk (approved):
LGTM 👍

PR #1097, opened 2026-06-01
FM-666 repoint api to cas2 hdc

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-666

Repoint CAS 2 HDC to use new `/cas2-hdc/` controller endpoints (these are a copy of the `/cas2/` endpoints, just renamed so we can free `/cas2/` to be used for bail + custody applications).

**N.B.** 
There is an upcoming API PR which removes `/cas2/` endpoints from the API see [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4806).

I have run the HDC and Bail end to end tests using this UI branch `feature/FM-666-repoint-API-to-cas2-hdc` and the upcoming API branch `feature/FM-704-remove-cas2-endpoints` which no longer contains `/cas2/` endpoints and both HDC and Bail e2e tests pass.

### DIFF
--- a/integration_tests/mockApis/applications.ts
+++ b/integration_tests/mockApis/applications.ts
@@ -12,7 +12,7 @@ export default {
     stubFor({
       request: {
         method: 'POST',
-        url: `/cas2/applications`,
+        url: `/cas2-hdc/applications`,
       },
       response: {
         status: 201,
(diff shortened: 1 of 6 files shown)


bobmeredith (approved):
Nice one!
It's very annoying that this code base uses literal paths for the mocks rather than the `paths` file. Still, it's not really worth fixing at this stage I guess.

PR #1102, opened 2026-06-10
FM-720 Update types and use them

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-720

This is part of wider tasks to rename CAS 2 to CAS 2 HDC so we can free up CAS 2 to be used for non HDC applications.

We have updated the types and controllers now in the API such that legacy CAS 2 data transfer objects and endpoints now reference Cas2Hdc instead of Cas2.

We can update the UI now to use the new names.

e2e tests have been run successfully locally using this branch of the UI and API main.

A subsequent PR will remove the aliasing from the OpenAPI config so that we can then update the types in the UI which will remove the unreferenced and legacy Cas2 types from the repository.  e2e tests have also successfully been run locally using an API branch with aliases removed and UI branch with legacy types removed to check this.

**Tips for reviewing:**

Commit 1 - add new `Cas2Hdc` types, set existing `Cas2` types to deprecated and export new type.  Basically a type update resulting from running `./generate-types`.
Commit 2 - use new `Cas2Hdc` types.  This is basically just changes to import statements to use new types.


integration_tests/pages/apply/before_you_start/check_information_needed/checkInformationNeededPage.ts
```diff
@@ -16,7 +16,7 @@ export default class CheckInformationNeededPage extends ApplyPage {
     )
   }
 
-  static visit = (application: Cas2Application) => {
+  static visit = (application: Cas2HdcApplication) => {
```
bobmeredith, line 19: Could just use the Alias `Application` here and clean up the import.

integration_tests/pages/apply/before_you_start/confirm_consent/confirmConsentPage.ts
```diff
@@ -16,7 +16,7 @@ export default class ConfirmConsentPage extends ApplyPage {
     )
   }
 
-  static visit = (application: Cas2Application) => {
+  static visit = (application: Cas2HdcApplication) => {
```
bobmeredith, line 19: Same as file above

integration_tests/pages/apply/before_you_start/referrer_details/confirmDetailsPage.ts
```diff
@@ -10,7 +10,7 @@ export default class ConfirmDetailsPage extends ApplyPage {
     super(`Confirm your details`, application, 'referrer-details', 'confirm-details')
   }
 
-  static visit = (application: Cas2Application) => {
+  static visit = (application: Cas2HdcApplication) => {
```
bobmeredith, line 13: Here's another one

integration_tests/pages/apply/before_you_start/referrer_details/contactNumberPage.ts
```diff
@@ -10,7 +10,7 @@ export default class ContactNumberPage extends ApplyPage {
     super(`What is your contact telephone number?`, application, 'referrer-details', 'contact-number')
   }
 
-  static visit = (application: Cas2Application) => {
+  static visit = (application: Cas2HdcApplication) => {
```
bobmeredith, line 13: OK, bored now!

integration_tests/pages/apply/before_you_start/referrer_details/jobTitlePage.ts
```diff
@@ -10,7 +10,7 @@ export default class JobTitlePage extends ApplyPage {
     super(`What is your job title?`, application, 'referrer-details', 'job-title')
   }
 
-  static visit = (application: Cas2Application) => {
+  static visit = (application: Cas2HdcApplication) => {
```
bobmeredith, line 13: Last one?

server/@types/shared/models/ApplicationOrigin.ts
```diff
@@ -2,4 +2,4 @@
 /* istanbul ignore file */
 /* tslint:disable */
 /* eslint-disable */
-export type ApplicationOrigin = 'courtBail' | 'prisonBail' | 'homeDetentionCurfew';
+export type ApplicationOrigin = 'courtBail' | 'prisonBail' | 'homeDetentionCurfew' | 'other';
```
bobmeredith, line 5: Is this correct in fact? Cas2/Cas2hdc will only ever be creating applications with an ApplicationOrigin of `homeDetentionCurfew` but perhaps it's not possible to limit it here.

bobmeredith (approved):
252 files!
Looks good,
Needless to say, it was more of a scan than a review 😄 

It shows the value of using aliases on imports! I did spot a few files where the import was aliased, but some code was using the non-aliased import as well which is a bit rubbish, but feel free to ignore.  I stopped highlighting them when I got bored!
The only thing of note is that the `ApplicationOrigin` type contains all the possible values, even though all-but-one are not applicable for Cas2-hdc. This probably can't be helped anyway TBH.

PR #1103, opened 2026-06-11
FM-710 Use OAuth Token for Snyk Jobs

sajifdeva-moj:
Updated Snyk Workflow to use OAuth tokens instead of API tokens.

Ran the SNYK workflow on this branch and it passed.


davidatkinsuk (approved):
LGTM 👍

PR #1108, opened 2026-06-24
Harden github actions and update renovate

davidatkinsuk:
* Use latest template renovate config
* Enable renovate dashboard
* Unpin workflows
* Harden use of actions/checkout
* Fix incorrect casing on github actions tags


daveawc (approved):
👍

